PR #6, opened 2016-03-08
Fixes inout error

ArtSabintsev:
This error appears when installing kitura via docker

```
/root/Kitura/Packages/Kitura-net-0.3.5/Sources/KituraNet/BufferList.swift:88:27: warning: 'inout' before a parameter name is deprecated, place it before the parameter type instead
    public func fillArray(inout buffer: [UInt8]) -> Int {
```

### DIFF
--- a/Sources/KituraNet/BufferList.swift
+++ b/Sources/KituraNet/BufferList.swift
@@ -85,7 +85,7 @@ public class BufferList {
     ///
     /// - Returns:
     ///
-    public func fillArray(inout buffer: [UInt8]) -> Int {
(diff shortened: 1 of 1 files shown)


dfirsht:
Hey @ArtSabintsev, this actually has already been fixed on the develop branch. You should see it in the docker image now. For future reference, you should submit any PRs to develop. In addition, before we can accept any PRs from you, you'll need to fill out a CLA as outlined here: https://github.com/IBM-Swift/Kitura/blob/master/CONTRIBUTING.md

ArtSabintsev:
Thanks for the update! I'll look at the CLA momentarily.

PR #12, opened 2016-03-15
Set SO_NOSIGPIPE on listening sockets to avoid dying from SIGPIPE on interrupted connections.

neotron:
Not sure if this is the best way to put it, but since the FD isn't exposed to the outer layer, it seemed like an appropriate location.

### DIFF
--- a/Sources/KituraNet/HttpServerSpi.swift
+++ b/Sources/KituraNet/HttpServerSpi.swift
@@ -14,6 +14,12 @@
  * limitations under the License.
  **/
 
+#if os(OSX) || os(iOS) || os(tvOS) || os(watchOS)
+	import Darwin
+#elseif os(Linux)
+	import Glibc
+#endif
+
 import BlueSocket
 import LoggerAPI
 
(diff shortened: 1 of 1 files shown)


billabt:
This is going to be in the next version of BlueSocket.

neotron:
Makes sense to keep out platform specific stuff from this layer.

kweinmeister:
Thanks for your PR! Please fill out the one time CLA form at https://github.com/IBM-Swift/CLA so we can accept it. Much appreciated.

billabt:
FYI, there's no `SO_NOSIGPIPE` on Linux.  To avoid the interruption, you need to add a `MSG_NOSIGNAL` flag to the underlying socket send() function.

neotron:
Yeah, I should have known that - have a C++ cross-platform piece of code that uses MSG_NOSIGNAL when available. I assume that this particular change should be closed for now, given that it'll happen in BlueSocket instead of here (and need x-platform change as well).

billabt:
Yup.  Thanks again for your efforts.  Much appreciated.

PR #13, opened 2016-03-15
Rename BlueSocketError to Error

complexityclass:


dfirsht:
Hey @complexityclass, I just changed master to only support BlueSockets v0.1 so it won't break. We'll be upgrading to 0.2 support in the develop branch of Kitura-net. Please resubmit your PR there. Also as a reminder, you need to fill out a CLA to contribute as specified here: https://github.com/IBM-Swift/Kitura/blob/develop/CONTRIBUTING.md

PR #46, opened 2016-07-12
IBM-Swift/Kitura#566 Change implementation of ClientRequest to be async

dfirsht:
Closes IBM-Swift/Kitura#566
## Checklist:


dfirsht:
I looked into the failures on Linux and found that it happened when two requests were trying to connect at the same time. This can be solved by changing the request queue to be serial rather than parallel. However I'm not sure if this is the behavior we want. @shmuelk thoughts?

dfirsht:
@shmuelk Please validate.

collinhundley:
I'm glad this is finally happening; can't wait to get it merged.

If you don't mind me asking a question here, because it involves `ClientRequest` and I don't want to open an issue for it:

How do we add an HTTP body to a `ClientRequest` for POST/PUT? I honestly can't figure out if that's even implemented right now. Thanks!

shmuelk:
@collinhundley You simply invoke ClientRequest's write functions before invoking the end function to send the data.

dfirsht:
Closing for now, we need to rework how to implement async in order to not negatively impact performance.

PR #51, opened 2016-07-20
Tuning and Cleanup of FastCGI Code

simplepanda:
PR to provide tuning and cleanup of FastCGI protocol implementation.

## Description
- Modified parser to assume HTTP/0.9 as the default, then parsing out actual protocol.
- Added detector for 0 length SERVER_PROTOCOL when rough HTTP/0.9 request arrives via FastCGI
- Simplified request URL processing down to simply URI use (same method as the HTTP server implementation) - bypasses potential host name / port parsing problems completely and removes associated processing overhead.
- Optimized logic where a block of data read from the socket would be returned from the FastCGI protocol parser as a zero length NSData() in order to signify the block was used completely. Now returns nil. Reduces NSData allocations non-trivially.
- Cleaned up code to removed stray self.X and ()'d if conditions (improved readability)

## Motivation and Context
Tuning and cleanup and overall code improvements.

## How Has This Been Tested?
Passes protocol test suite. Also load tested on both OS X and Linux with expected results.

## Checklist:
- [x] If applicable, I have added tests to cover my changes (tests updated)

### DIFF
--- a/Sources/KituraNet/FastCGIRecordCreate.swift
+++ b/Sources/KituraNet/FastCGIRecordCreate.swift
@@ -78,7 +78,7 @@ class FastCGIRecordCreate {
         let r = NSMutableData()
         
         var v : UInt8 = FastCGI.Constants.FASTCGI_PROTOCOL_VERSION
-        var t : UInt8 = self.recordType
+        var t : UInt8 = recordType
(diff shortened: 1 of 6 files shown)


simplepanda:
Thanks @ianpartridge - can you perhaps tag a 0.22.2 on this now that it's merged?

ianpartridge:
Done.

PR #54, opened 2016-07-24
An initial keep alive implementation

shmuelk:
A working initial implementation of keep alive and asynchronous socket I/O.

## Description
This is the result of the work in issue IBM-Swift/Kitura#435. We now have an initial working implementation of keep alive for both Linux and OSX. The Linux version uses epoll and the OSX version uses DispatchSource to be notified of waiting data to be read.

## Motivation and Context
Improved performance, especially on OSX in various performance tests.

## How Has This Been Tested?
The TechEmpower performance test has been run along with an additional test added to the normal KituraNet test bucket.

## Checklist:


Tests/LinuxMain.swift
```diff
@@ -21,6 +21,7 @@ import XCTest
 XCTMain([
 	testCase(ClientE2ETests.allTests),
   	testCase(ClientRequestTests.allTests),
+        testCase(LargePayloadTests.allTests),
```
vadimeisenbergibm: Redundant whitespace

Sources/KituraNet/HTTPIncomingMessage.swift
```diff
@@ -223,6 +226,57 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
             callback(.success)
         }
     }
+
```
vadimeisenbergibm: It looks like code duplication of `parse(HTTPParserErrorType)`. Can `parse(HTTPParserErrorType)` call `parse(NSData)` passing it `ioBuffer` ?

Sources/KituraNet/HTTPIncomingMessage.swift
```diff
@@ -294,7 +347,7 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
     func drain() {
         if let parser = httpParser {
             saveBody = false
-            while status == .headersComplete {
+            while status == .headersComplete  ||  status == .headersCompleteKeepAlive {
```
vadimeisenbergibm: It seems there is a confusion here between Parser State and Parser Status. Parser State designates which state we are currently in: we started parsing, we parsed headers, we parsed message. On top of that there is a status - which state we are in (initial, after headers, after message), reset was asked, error happened, keepAlive was read. I would create `HTTParserStatus` that would contain `HTTPParserState`, boolean `keepAliveFlag`, and `HTTPParserErrorType`. I am not sure where `reset` should go.

With parser state vs. parser status logic there is no need to check for both `headersComplete` and `headersCompleteKeepAlive`. The code would become: `while status.state == .headersComplete`.

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,166 @@
...
+                let event = pollingEvents[index]
+                
+                if  (event.events & EPOLLERR.rawValue)  == 1  ||  (event.events & EPOLLHUP.rawValue) == 1  ||  (event.events & EPOLLIN.rawValue) == 0 {
+                    
+                    print("Error occurred on a file descriptor of an epool wait")
+                    
+                }
+                else {
+                    if  let handler = socketHandlers[event.data.fd] {
+                        handler.handleRead()
```
vadimeisenbergibm: So the accepted sockets will not participate in epoll?

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,166 @@
...
+        let now = NSDate()
+        guard  now.timeIntervalSince(keepAliveIdleLastTimeChecked) > keepAliveIdleCheckingInterval  else { return }
+        
+        let maxInterval = now.timeIntervalSinceReferenceDate
+        print("Checking for idle sockets to close")
+        for (fileDescriptor, handler) in socketHandlers {
+            if  handler.inProgress  ||  maxInterval < handler.keepAliveUntil {
+                continue
+            }
+            print("closing idle socket \(fileDescriptor)")
```
vadimeisenbergibm: Log should be used

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,166 @@
...
+            socketHandlers.removeValue(forKey: fileDescriptor)
+                
+            let result = epoll_ctl(epollDescriptor, EPOLL_CTL_DEL, fileDescriptor, nil)
+            if  result == -1  {
+                Log.error("epoll_ctl failure. Error code=\(errno). Reason=\(lastError())")
+                print("epoll_ctl failure. Error code=\(errno). Reason=\(lastError())")
```
vadimeisenbergibm: Log should be used instead of print

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Correct.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Oops. Those were debugging statements. Let me remove them.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Those are also debug statements and should be removed.

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,166 @@
...
+            if  count == 0  {
+                continue
+            }
+            
+            for  index in 0  ..< count {
+                let event = pollingEvents[index]
+                
+                if  (event.events & EPOLLERR.rawValue)  == 1  ||  (event.events & EPOLLHUP.rawValue) == 1  ||  (event.events & EPOLLIN.rawValue) == 0 {
+                    
+                    print("Error occurred on a file descriptor of an epool wait")
```
vadimeisenbergibm: Log should be used instead of print

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,166 @@
...
+                    
+                }
+                else {
+                    if  let handler = socketHandlers[event.data.fd] {
+                        handler.handleRead()
+                    }
+                    else {
+                        print("No handler for file descriptor \(event.data.fd)")
```
vadimeisenbergibm: Log should be used instead of print

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Fixed.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Fixed

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: However, those print statements should be Log.error calls.

This has been fixed.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Correct and fixed.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Correct and fixed.

Sources/KituraNet/IncomingSocketHandler.swift
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+// MARK: IncomingSocketHandler protocol
+
+import Foundation
+
+
+protocol IncomingSocketHandler {
```
vadimeisenbergibm: It seems `IncomingSocketHandler` is not used and `IncomingHTTPSocketHandler` is used instead. I would create `IncomingHTTPSocketHandler` protocol that would extend (inherit from) `IncomingSocketHandler`, and rename `IncomingHTTPSocketHandler`  class into something like `BasicIncomingHTTPSocketHandler`.

Tests/LinuxMain.swift
shmuelk: Fixed

Sources/KituraNet/IncomingSocketHandler.swift
shmuelk: There was originally a plan for things like IncomingWebSocketHandler, which would also implement IncomingSocketHandler.

The more I thought about it and experienced various issues with modifying Dictionary's in multiple threads, I have come to the conclusion that what really needs to be done is to eliminate the IncomingSocketHandler protocol, rename IncomingHTTPSocketHandler to IncomingSocketHandler, and remove all HTTP handling from the resulting IncomingSocketHandler class. That class would be only responsible for the I/O. It would get a reference to a processor class that would implement a common protocol, which would be independent of whether or not it was HTTP or WebSockets or something else.....

Sources/KituraNet/IncomingHTTPSocketHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+        request = HTTPServerRequest(reader: reader)
+        
+        response = HTTPServerResponse(handler: self)
+        
+        // Perform any platform specific initialization
+        setup()
+    }
+    
+    ///
+    /// Read in the available data and hand off to ommon processing code
```
vadimeisenbergibm: Typo: _"ommon"_

Sources/KituraNet/HTTPIncomingMessage.swift
shmuelk: Probably. At one time I wanted to remove `parse(HTTPParserErrorType)`.  It is only called from ClientRequest at this time. @dfirsht started to work on making that code asynchronous and didn't want to make a mess at this time.

Sources/KituraNet/PseudoAsynchronousReader.swift
```diff
@@ -0,0 +1,124 @@
...
+    init(clientSocket: Socket) {
+        self.clientSocket = clientSocket
+        
+        #if os(Linux)
+            waitingForInput = dispatch_semaphore_create(0)
+            readBufferLock = dispatch_semaphore_create(1)
+        #endif
+    }
+    
+    func addToAvailableData(from: NSData) {
```
vadimeisenbergibm: Comments that describe the algorithm and the assumptions are required. It seems we have a case of Producer-Consumer pattern using Semaphores - https://en.wikipedia.org/wiki/Producer–consumer_problem#Using_semaphores .

`addAvailableData` is `produce()`, `readSynchronously` is `consume`, `waitingForInput` is `consume` semaphore, `readBufferLock` is `produce` semaphore. Consider to change the names according to the pattern. `Lock` is a confusing name, it is not a lock but a semaphore. 

If so, it is not clear why `addToAvailableData` should signal to `readBufferLock`, and why `readSynchronously` should signal to `waitingForInput`. Shouldn't each function wait for exactly one semaphore and signal the other one? 

Could there be multiple producers/consumers? it should be documented.

Sources/KituraNet/IncomingHTTPSocketHandler.swift
```diff
@@ -0,0 +1,241 @@
...
+    ///
+    /// Should this socket actually be kep alive?
+    ///
+    var isKeepAlive: Bool { return clientRequestedKeepAlive && numberOfRequests > 0 }
+    
+    ///
+    /// An enum for internal state
+    ///
+    enum State {
+        case reset, initial, parsed
```
vadimeisenbergibm: Should be `parsedHeaders` instead of `parsed`, otherwise the reader can think that the whole message, including the body is parsed.

Sources/KituraNet/PseudoAsynchronousReader.swift
```diff
@@ -0,0 +1,124 @@
...
+ * limitations under the License.
+ **/
+
+import Dispatch
+import Foundation
+
+import LoggerAPI
+import Socket
+
+class PseudoAsynchronousReader {
```
vadimeisenbergibm: I am confused why the class is called `PseudoAsynchronousReader`. It is indeed asynchronous, so maybe you meant `PseudoSynchronousReader`? How about `ProducerConsumerBuffer` with `produce` and `consume` methods?

Sources/KituraNet/IncomingHTTPSocketHandler.swift
shmuelk: Fixed

Sources/KituraNet/IncomingHTTPSocketHandler.swift
shmuelk: Fixed.

Sources/KituraNet/HTTPIncomingMessage.swift
shmuelk: Fixed.

Sources/KituraNet/HTTPIncomingMessage.swift
shmuelk: The extra parse function has been removed.

Sources/KituraNet/IncomingHTTPSocketHandler.swift
```diff
@@ -0,0 +1,204 @@
...
+    
+    /// Write data to the socket
+    func write(from data: NSData) {
+        guard socket.socketfd > -1  else { return }
+        
+        do {
+            try socket.write(from: data)
+        }
+        catch {
+            print("Write to socket (file descriptor \(socket.socketfd) failed. Error number=\(errno). Message=\(errorString(error: errno)).")
```
ianpartridge: Remove?

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,145 @@
...
+    private var socketHandlers = [Int32: IncomingHTTPSocketHandler]()
+    
+    /// Interval at which to check for idle sockets to close
+    let keepAliveIdleCheckingInterval: NSTimeInterval = 60.0
+    
+    /// The last time we checked for an idle socket
+    var keepAliveIdleLastTimeChecked = NSDate()
+    
+    init() {
+        epollDescriptor = epoll_create1(0)
```
ianpartridge: Why not `epoll_create()` - comment needed here.

Sources/KituraNet/OSXIncomingSocketManager.swift
```diff
@@ -0,0 +1,90 @@
...
+    func handle(socket: Socket, using: ServerDelegate) {
+        
+        do {
+            try socket.setBlocking(mode: false)
+            
+            let handler = IncomingHTTPSocketHandler(socket: socket, using: using)
+            socketHandlers[socket.socketfd] = handler
+        }
+        catch {
+            print("Failed to make incoming socket (File Descriptor=\(socket.socketfd)) non-blocking. Error code=\(errno). Reason=\(lastError())")
```
ianpartridge: Remove?

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,145 @@
...
+            let result = epoll_ctl(epollDescriptor, EPOLL_CTL_ADD, handler.fileDescriptor, &event)
+            if  result == -1  {
+                Log.error("epoll_ctl failure. Error code=\(errno). Reason=\(lastError())")
+            }
+        }
+        catch {
+            Log.error("Failed to make incoming socket (File Descriptor=\(socket.socketfd)) non-blocking. Error code=\(errno). Reason=\(lastError())")
+        }
+        
+        removeIdleSockets()
```
ianpartridge: Does this mean idle sockets are only pruned if new connections keep arriving?  Would it be better to have a separate thread running with a timer to do this periodically?

Sources/KituraNet/HTTPServerResponse.swift
```diff
@@ -15,18 +15,12 @@
...
 
 // MARK: HTTPServerResponse
 
-public class HTTPServerResponse : SocketWriter, ServerResponse {
-
-    ///
-    /// Socket for the HTTPServerResponse
-    ///
-    private var socket: Socket?
+public class HTTPServerResponse : ServerResponse {
```
ianpartridge: Comments needed here - what is an `HTTPServerResponse` and what is it for.

Sources/KituraNet/IncomingSocketHandler.swift
ianpartridge: This seems an important new protocol - comment/documentation needed.

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,145 @@
...
+
+import Glibc
+import Foundation
+
+import CEpoll
+import KituraSys
+import LoggerAPI
+import Socket
+
+class IncomingSocketManager  {
```
ianpartridge: Comments needed here.

Sources/KituraNet/LinuxIncomingSocketManager.swift
```diff
@@ -0,0 +1,145 @@
...
+                    }
+                    else {
+                        Log.error("No handler for file descriptor \(event.data.fd)")
+                    }
+                }
+            }
+        }
+    }
+    
+    /// Remove idle sockets
```
ianpartridge: Useless comment 😉 - sorry but it is!

Sources/KituraNet/OSXIncomingSocketManager.swift
```diff
@@ -0,0 +1,90 @@
...
+ **/
+
+#if os(OSX) || os(iOS) || os(tvOS) || os(watchOS)
+
+import Foundation
+   
+import LoggerAPI
+import Socket
+    
+class IncomingSocketManager  {
```
ianpartridge: Documentation needed for this new class.

ianpartridge:
@shmuelk this is quite hard to review without an overview of the changes from an architectural point of view. But I've made a few comments 😃

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: I thought that if no new sockets were coming in, it wouldn't matter quite so much if we didn't clean them up right away.

But perhaps you're right.

ianpartridge: Perhaps you're right! Just wanted to check my understanding was correct.

Sources/KituraNet/OSXIncomingSocketManager.swift
shmuelk: Fixed

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Changed to use epoll_create

Sources/KituraNet/IncomingHTTPSocketHandler.swift
shmuelk: Removed.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -258,15 +219,14 @@ public class ClientRequest: SocketWriter {
...
-                                print("ClientRequest error. Failed to parse response. status=\(status)")
-                        }
+                    let parseStatus = response.parse()
+                    switch(parseStatus.error) {
+                        case .success:
+                            self.callback(response: self.response)
+                            callCallback = false
+
+                        default:
+                            print("ClientRequest error. Failed to parse response. status=\(parseStatus.error)")
```
ianpartridge: One more print() here that should be Logger?

shmuelk: Done.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Comments added to the code indicating this.

djones6:
@shmuelk re. the two implementations of `IncomingSocketManager`:

One thing I would like to continue to do is to track the performance delta between the 'OSX' (cross-platform) and 'Linux' (epoll) implementations when running on Linux.

A couple of concerns:

1. I would want us to ensure that the cross-platform implementation will always build on Linux.  At the moment, we won't attempt to compile that code on Linux and so breakage wouldn't be detected.

2. It would be nice if there were a more convenient way of switching between implementations (such as changing one if block in the code).  At the moment I'd need to remove the file containing the Linux implementation and the if guard in the x-platform version.  That said, I imagine a suitable git patch could be generated.

shmuelk:
@djones6 I plan to do a second pass of refactoring. I will try to see what can be done for you.

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Done.

shmuelk:
@vadimeisenbergibm I guess we can drop .success and leave the just details on the type of error in the HTTPParserErrorType enum.

Sources/KituraNet/PseudoAsynchronousReader.swift
shmuelk: I'd like to rename `PseudoAsynchronousReader` to `PseudoSynchronousReader` and then rename two of it's functions as follows:
   - `addToAvailableData` to `append`
   - `readSynchronously` to `read`

Does that make it more understandable?

vadimeisenbergibm: How about these names: `addDataToRead` and `read`? This will stress the point that these are _matching_ functions.

shmuelk: ok

shmuelk:
I see. You want me to take out the state of error and instead make the error field an optional. It will only be set if there was an error. That's the pattern you want?

vadimeisenbergibm:
yes.

Sources/KituraNet/HTTPServerResponse.swift
shmuelk: Done

Sources/KituraNet/OSXIncomingSocketManager.swift
shmuelk: Done

Sources/KituraNet/LinuxIncomingSocketManager.swift
shmuelk: Improved comments added.

Sources/KituraNet/IncomingSocketHandler.swift
shmuelk: I've added comments to the class implementing this protocol. I plan a refactoring hopefully in a day or two that will eliminate the IncomingSocketHandler protocol. The IncomingHTTPSocketHandler class will be renamed to IncomingSocketHandler and the HTTP specific parts of it will be moved to a new class.

shmuelk:
Done

Sources/KituraNet/IncomingHTTPSocketHandler.swift
```diff
@@ -0,0 +1,202 @@
...
+            
+        case .initial:
+            inProgress = true
+            HTTPServer.clientHandlerQueue.enqueueAsynchronously() { [unowned self] in
+                self.parse(buffer)
+            }
+            
+        case .parsedHeaders:
+            reader.addDataToRead(from: buffer)
+            break
```
dfirsht: break not needed here

Issue #52, opened 2016-07-20
Swap HTTPParser Swift package for http-parser


dsperling:
We have been working a pure Swift port of the 'C' http-parser project.

https://github.com/smithmicro/HTTPParser

Swapping HTTPParser in for http-parser would get Kitura closer to pure Swift.  Please review the performance of HTTPParser in the readme to see if it would meet the requirements of Kitura.

I will work on a PR, but I wanted to alert you to this new project and gauge interest.

ianpartridge:
Hi @dsperling this sounds interesting. What was your reason for developing it? The Node.js HTTP parser is mature and well-used.

rfdickerson:
Yes! Super excited about this.

1) It will make it easier to work on a Kitura application with XCode

  Right now we need to add $SRCROOT/.build/debug to the Library search paths for KituraNet and Kitura

2) Right now whenever upstream http-parser gets updated, we have to copy the c and h file to our CHttpParser library.

ianpartridge:
Good points @rfdickerson .

@dsperling if you can open a PR we can get the new code benched to see if there's a performance impact.

dsperling:
@ianpartridge We started developing the Swift port initially as a thought experiment.   We wanted to see how Swift 3 was performing in comparison to 'C' code and selected http-parser as our test project.  The port went quickly and we were pleased with the performance.

There is still plenty to do.  The unit tests still need work and will be 2-3 times the effort of porting the library itself.  Hopefully we can recruit some of the original authors to help out.

I will get a PR in soon.

ianpartridge:
@dsperling cool 👍 

You might be interested in this article I read a while ago - cool work: http://www.cocoawithlove.com/blog/2016/05/01/swift-name-demangling.html

rfdickerson:
That's very exciting work, @dsperling! Once you get a PR in, I'm curious if we can do some benchmark analysis here. Maybe @djones6 could get involved with the performance testing.

ianpartridge:
@rfdickerson he's lined up 😉

dsperling:
I have completed the integration of HTTPParser, but I cannot submit a PR because of a Linux linker issue.  I am wondering if you guys have seen this.

Since we now have NSURL on Linux, we did not port the URL parsing of http-parser.  So with this change:
* HTTPParser/HTTPParser.swift uses HTTPParser
* HTTPParser/URLParser.swift uses NSURL

[...]

PR #55, opened 2016-07-26
Swap HTTPParser Swift package for http-parser - issue #52

dsperling:
## Description
Swapped Swift HTTPParser for 'C' http-parser ("CHTTPParser").  HTTPParser is a direct port of http-parser with the goal of being as maintainable as possible.

## Motivation and Context
This was initially a thought experiment, but the performance of HTTPParser was only 1.4 times slower than http-parser.  It processes over 500K requests per second using the 'bench' tool in Release mode.  Performance tuning continues.

We also noticed that the CHTTPParser module was causing Xcode compile issues for new Kitura developers.

## To Do
* Performance comparisons between Kitura based apps using HTTPParser vs CHTTPParser
* Test use of NSData bytesNoCopy inside of HTTPParser callbacks within Kitura-net.

## How Has This Been Tested?
Kitura-net unit tests pass on Linux and Mac
Kitura unit tests pass on Linux and Mac

## Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,6 +22,6 @@ let package = Package(
         .Package(url: "https://github.com/IBM-Swift/Kitura-sys.git", majorVersion: 0, minor: 22),
         .Package(url: "https://github.com/IBM-Swift/BlueSocket.git", majorVersion: 0, minor: 6),
         .Package(url: "https://github.com/IBM-Swift/CCurl.git", majorVersion: 0, minor: 2),
-        .Package(url: "https://github.com/IBM-Swift/CHTTPParser.git", majorVersion: 0, minor: 2),
+        .Package(url: "https://github.com/smithmicro/HTTPParser.git", majorVersion: 2, minor: 7)
     ]
 )
--- a/Sources/KituraNet/HTTPParser/HTTPParser.swift
+++ b/Sources/KituraNet/HTTPParser/HTTPParser.swift
@@ -15,22 +15,17 @@
  **/
 
 import KituraSys
-import CHTTPParser
+import HTTPParser
 import Foundation
 
 // MARK: HTTPParser
 
-class HTTPParser {
+class HTTPParser : http_parser_delegate {
 
     ///
     /// A Handle to the HTTPParser C-library
     ///
     var parser: http_parser
-
-    ///
-    /// Settings used for HTTPParser
-    ///
-    var settings: http_parser_settings
 
     ///
     /// Parsing a request? (or a response)
@@ -40,24 +35,8 @@ class HTTPParser {
     ///
     /// Delegate used for the parsing
     ///
-    var delegate: HTTPParserDelegate? {
-
-        didSet {
-            if let _ = delegate {
-                withUnsafeMutablePointer(&delegate) {
-                    ptr in
-                    self.parser.data = UnsafeMutablePointer<Void>(ptr)
-                }
-            }
-        }
-        
-    }
+    var delegate: HTTPParserDelegate?
     
-    ///
-    /// Whether to upgrade the HTTP connection to HTTP 1.1
-    ///
-    var upgrade = 1
-
     ///
     /// Initializes a HTTPParser instance
     ///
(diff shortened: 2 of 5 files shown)


ianpartridge:
@shmuelk can you review this please?

@dsperling we will probably not want to depend on a 3rd party repo - would you be ok with us forking your repo to IBM-Swift/HTTPParser and making Kitura-net depend on that? We would sync our HTTPParser repo with smithmicro regularly and all development work would happen in your repo. Thoughts?

dsperling:
@ianpartridge I don't have a problem with forking.  I'm glad the port is of interest.

However I would like to understand the reasoning behind your 3rd party repro concerns.  My assumption was that the Swift Package Manager would do the heavy lifting for us where Kitura-net would be pinned to specific HTTPParser releases.  It seems like a extra step having a intermediary repro that needs to be synced.

Another option is that we move the project to IBM-Swift sandbox and I simply provide PR's.  Either way works for us.

ianpartridge:
Thanks @dsperling. The slight concern about depending on a 3rd party repo is that if a git tag was deleted for some reason, Kitura would stop building. Unlikely I know, but... We have a large and growing stack of projects and repos which rely on Kitura-net.

The offer to host this under IBM-Swift is very much appreciated. I'll discuss with our team and get back to you if that's OK?

dsperling:
Sounds good @ianpartridge

djones6:
Just ran some performance numbers against this, patched onto Kitura 0.23:

### Linux
```
Implementation | Avg Throughput | Max Throughput | Avg CPU | Avg RSS (kb) | Avg Lat (ms) | Max Lat (ms)
---------------|----------------|----------------|---------|--------------|--------------|--------------
          0.23 |        51837.1 |       52991.06 |    90.6 |       537110 |          2.6 |       219.73
     0.23+pr55 |        45890.1 |       46424.85 |    89.7 |       682325 |          3.0 |        209.5
```
Throughput reduction is around 12%.  I'll grab a profile from Perf later on.
### Mac
```
Implementation | Avg Throughput | Max Throughput | Avg CPU | Avg RSS (kb) | Avg Lat (ms) | Max Lat (ms) 
---------------|----------------|----------------|---------|--------------|--------------|--------------
          0.23 |        18602.5 |       18937.34 |    77.3 |       419451 |          7.3 |       120.37 
     0.23+pr55 |        18539.1 |       18850.79 |    78.2 |       490662 |          8.5 |       452.95 
```
Doesn't seem to significantly impact performance on Mac (throughput + cpu differences are within noise), however we have scalability issues at the moment which could obscure a simple compare, we'd need to look at a profile to confirm where time is really being spent.

djones6:
Profiles for the compare on Linux:

[perf-report.023.txt](https://github.com/IBM-Swift/Kitura-net/files/388415/perf-report.023.txt)
Requests/sec:  50386.92
Average CPU util: 91.8 %

[perf-report.023pr55.txt](https://github.com/IBM-Swift/Kitura-net/files/388416/perf-report.023pr55.txt)
Requests/sec:  44668.58
Average CPU util: 91.0 %

dsperling:
This is consistent with what I see using the http-parser stress code.  Time spent in the core parser is:
- 1.51%  HTTPParser.http_parser.execute
- 1.19%  http_parser_execute
- 27% slower

With the ported stress.c to swift, I see:
- http-parser - 'C' - 782K req/sec
- HTTPParser - Swift - 566K req/sec
- 38% slower

While HTTPParser does not add memory allocation overhead, it is still adding overhead via retain/release calls.  I am assuing that is reflected here:

023:
- 0.35%  TechEmpower  libswiftCore.so      [.] _swift_release_n_
- 0.34%  TechEmpower  libswiftCore.so      [.] _swift_retain_n_

023pr55:
- 0.70%  TechEmpower  libswiftCore.so      [.] _swift_retain_n_
- 0.70%  TechEmpower  libswiftCore.so      [.] _swift_release_n_

This should be the focus on further HTTPParser optimization.  Are we "good enough" to proceed?  What does the team think?

One risk item is that not all unit tests for HTTPParser are ported.  This is progressing but at a slow pace.

ianpartridge:
Hi @dsperling - I don't think we could merge a PR which impacts Kitura throughput by this much, unfortunately. We are definitely interested in reducing our dependency on external C code, but HTTP parsing is a sensitive area so we'll need to ensure quality (through the ported testsuite) and performance (I'm not sure exactly what cost would be acceptable, I'd guess low single figures).

On the performance front, do you think it would be possible to reduce the ARC overhead? For example, by identifying hot loops that contain *swift_retain/release* calls and working around them. Just an idea?

dsperling:
@ianpartridge Thanks for your consideration.  We will continue to work on optimizations and test coverage.  I will close the PR and will report status via https://github.com/IBM-Swift/Kitura-net/issues/52.

I would ask that the team suggests an acceptable performance benchmark.  Right now we are 38% slower than http-parser based on bench.c performance tests.  You mentioned low single figures.  That does that mean < 5% slower?

ianpartridge:
Hi @dsperling!

The bench.c numbers you report seem to be worse than the Kitura throughput numbers that @djones6 measured - 38% compared to 12%. It's the Kitura performance that we're mainly interested in (I guess improving the bench.c results will help Kitura).

There's a tradeoff here: the win of using pure Swift code, ridding ourselves of a C dependency, and simplifying our build process; versus a potential loss of performance and the risks of switching to relatively untested code (the node.js parser has been field-tested for years).

So it's going to come down to a judgement call, eventually. It's fair to ask what the "bar" is though - I will discuss and see what other people's views are. What do you think an appropriate number would be?

As a matter of general principle Swift shouldn't be necessarily slower than C, as a compiled language. If it looks like ARC is the cost here, then it might be worth experimenting with using structs instead of classes as structs don't participate in ARC.

djones6:
To weigh in here, I agree that < 5% impact on Kitura is a good number to set the bar at in this instance.
In theory, a pure Swift implementation should give the Swift compiler more opportunities to optimize (if I understand correctly that calling into C code is an optimization boundary), so there's the potential for some future performance gain.

There are other considerations besides performance - future enhancement / maintenance, code quality / test coverage, 'swiftyness' - which it would be good to discuss.

gottesmm:
@ianpartridge Is it possible for you to submit some form of this as a stripped down benchmark upstream? We are interested in it as an optimization target.

ianpartridge:
Hi @gottesmm. This PR was for switching Kitura to use a Swift-based HTTP parser - it doesn't contain the new parser itself.

The new parser is at https://github.com/smithmicro/HTTPParser and was written by @dsperling as a port of the Node.js http_parser C code.

Perhaps @dsperling would be interested in contributing his code upstream?

dsperling:
@gottesmm - I would be happy to help.

I stopped work on the Swift HTTPParser back in September when it was clear that the codebase could not meet the performance requirements of Kitura in the short term.  The codebase if functional, but only 10% of the unit tests are ported.

I will get this upgraded to Swift 3.0.1.  I think using the Swift HTTPParser as a benchmark would be a great initial use of this port.

Question:  The codebase has a MIT license where Kitura is Apache 2.  Is that an issue?

Let me know how I can help strip this down for a benchmark.

gottesmm:
Ideally we would contribute this to the swift benchmark suite. That would require the code to licensed under the same license as Swift (which is Apache2). Is that possible?

In terms of stripping down the benchmark, I imagine you would want to create a library that could parse any HTTP header. Then we would write a test case that would invoke the HTTP header parser with the same raw pointer that is taken from a constant string.

What do you think?

dsperling:
@gottesmm The project is a port of https://github.com/nodejs/http-parser so I will need to contact the authors.  I will post a question in that project and tag you.

gottesmm:
Thanks!

dsperling:
@gottesmm - I couldn't tag you over there, so here is the issue.  Please add any color or correct any of my assumptions.

https://github.com/nodejs/http-parser/issues/344

PR #57, opened 2016-07-28
Resolve Issue IBM-Swift/Kitura#640

shmuelk:
Enable on Linux to use DispatchSource instead of epoll via a compile time switch

## Description
DispatchSource related code for keep alive is now selected if on MacOS and on Linux if the compile time switch GCD_ASYNCH is specified.

## Motivation and Context
This resolves IBM-Swift/Kitura#640

## Checklist:


Sources/KituraNet/IncomingSocketManager.swift
```diff
@@ -16,46 +16,60 @@
...
     
     /// Interval at which to check for idle sockets to close
-    let keepAliveIdleCheckingInterval: NSTimeInterval = 60.0
+    let keepAliveIdleCheckingInterval: TimeIntervalType = 60.0
     
     /// The last time we checked for an idle socket
-    var keepAliveIdleLastTimeChecked = NSDate()
+    var keepAliveIdleLastTimeChecked = DateType()
+    
+    #if GCD_ASYNCH
```
ianpartridge: These two lines can just be `#if os(Linux)`

ianpartridge: Ignore more, I'm wrong :)

ianpartridge: `#if os(Linux) && !GCD_ASYNCH` might be clearer though.

Sources/KituraNet/HTTP/IncomingHTTPSocketHandler.swift
```diff
@@ -135,13 +168,36 @@ class IncomingHTTPSocketHandler: IncomingSocketHandler {
...
+            source.cancel()
+        #elseif GCD_ASYNCH
+	    dispatch_source_cancel(source!)
+        #else
+            dispatchSourceCancelHandler()
+        #endif
+    }
+    
+    /// DispatchSource cancel handler
+    private func dispatchSourceCancelHandler() {
```
ianpartridge: I think this should be `handleCancel()`, to match with `handleRead()`?

Sources/KituraNet/IncomingSocketManager.swift
```diff
@@ -16,46 +16,60 @@
 
 
 #if os(Linux)
+    import Glibc
+    import CEpoll
```
ianpartridge: We don't need this if GCD_ASYNCH is set.

Sources/KituraNet/IncomingSocketManager.swift
shmuelk: It would if you could.....

PR #62, opened 2016-08-03
Fixed: SPIUtils.httpDate(_) left a visible Optional() on macOS

gormster:
SPIUtils.httpDate(_) would give output containing an Optional() around the year. For example,

```swift
let date = Date(timeIntervalSince1970:1249648496)
print(SPIUtils.httpDate(date))
```

prints the string `Fri, 07 Aug Optional(2009) 12:34:56 GMT`. This is obviously not a valid HTTP date.

## Description

These changes simply unwrap the optional into a new variable called `year` and use that in the format string instead.

## Motivation and Context

Kitura uses httpDate to set the value for a cookie's expiration, which was not a valid HTTP date.

## How Has This Been Tested?

A test case has been added to test the new code and ensure that httpDate(_) returns a valid HTTP date.

Testing has been performed on macOS, since that was the origin of the bug. To be honest, I'm not totally sure *how* to test on Linux, as I don't have a Linux box at my disposal. I'm sure there's some kind of magical Docker thing I could do..?

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
+++ b/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
@@ -121,6 +121,7 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
         if  status.state == .reset  {
             status.reset()
             parser.reset()
+            headers.removeAll()
         }
         
         var start = 0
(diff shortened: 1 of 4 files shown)


gormster:
whoops opened this from the wrong branch

PR #64, opened 2016-08-03
IBM-Swift/Kitura#667 Correct the way HTTPIncomingMessage is reset

shmuelk:
Properly reset a HTTPServerRequest and/or ClientResponse when they are "reused".

## Description
When a connection is reused, the HTTPServerRequest object is reused for the following request. Similarly when parsing the response to a ClientRequest and a Continue status code was found, the ClientResponse object is reused.

In both cases the reset was incorrectly done, in the shared super-class HTTPIncomingMessage.

## Motivation and Context
Resolves issue IBM-Swift/Kitura#667

## How Has This Been Tested?
Ran the test case written the issue reporter as well a the TechPower test.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
+++ b/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
@@ -119,8 +119,7 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
         
         // If we were reset because of keep alive
         if  status.state == .reset  {
-            status.reset()
-            parser.reset()
+            reallyReset()
         }
         
         var start = 0
@@ -134,9 +133,8 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
                 
                 if  status.state == .reset  {
                     // Apparently the short message was a Continue. Let's just keep on parsing
-                    status.state = .initial
                     start = numberParsed
-                    parser.reset()
+                    reallyReset()
                 }
                 else {
                     /* Handle error. Usually just close the connection. */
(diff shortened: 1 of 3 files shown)


Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
```diff
@@ -372,13 +370,20 @@ public class HTTPIncomingMessage : HTTPParserDelegate, SocketReader {
...
-    /// instructions for when reading is reset
+    /// Signal that reading is being reset
     func reset() {
-        lastHeaderWasAValue = false
-        url.length = 0
         status.state = .reset
     }
 
+    /// When we're ready, really reset everything
+    private func reallyReset() {
```
ianpartridge: I hate this name 😄 . Could it just be `reset()`, and we rename the function in `HTTPParserDelegate` to something else?

shmuelk: Actually the one to rename is the one right above this one.

Would you prefer `prepareToReset` and `reset`?

ianpartridge: I thought the one right above was named `reset()` because it was the name required by `HTTPParserDelegate` conformance?

shmuelk: That can also be fixed if needed....

ianpartridge: That was my point :)

PR #71, opened 2016-08-15
swiftlint autocorrect

rob-deans:
## Description
Swiftlint's autocorrect feature mainly reformatting the code 

## Checklist:

### DIFF
--- a/Sources/KituraNet/BufferList.swift
+++ b/Sources/KituraNet/BufferList.swift
@@ -22,12 +22,12 @@
 
 import Foundation
 
-// MARK: BufferList 
+// MARK: BufferList
 
 public class BufferList {
 
-    // MARK: -- Private 
-    
+    // MARK: -- Private
+
     ///
     /// Internal storage buffer
     ///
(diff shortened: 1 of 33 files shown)


rob-deans:
@shmuelk @ianpartridge @vadimeisenbergibm this is what happens when you run autocorrect with all the auto correctable rules enabled

ianpartridge:
It all looks sensible, I support this PR. It doesn't catch two spaces in `if  let` - we should open a bug against SwiftLint for that?

ianpartridge:
Also, let's look at https://houndci.com/ for automating this stuff?

shmuelk:
I disagree. I don't like what it does with if else.

Here is an example:

-        }
-        else if name.hasPrefix("HTTP_") {
-            
+
+        } else if name.hasPrefix("HTTP_") { 

It is placing the else on the same line as the closing brace of the if.

vadimeisenbergibm:
What is wrong with that? Note that in the original code above other `else if` statements are on the line of the closing brace. These are our Code Conventions, see https://github.com/raywenderlich/swift-style-guide#spacing, https://github.com/apple/swift-package-manager/blob/1bfda9bbdb963e11c8144c89b6b260106d38be6d/Sources/Utility/PkgConfig.swift#L139 .

PR #88, opened 2016-08-30
URL query parameters parsing

naithar:
Implements parsing of dictionaries and arrays passed in url query.

This PR is needed for successful build of IBM-Swift/Kitura#707 
## Motivation and Context

IBM-Swift/Kitura#691
## How Has This Been Tested?

Added tests to `ParserTests.swift` that test parameter parsing.
## Checklist:


shmuelk:
@naithar Thanks for the PR. Some folks our out on vacation this week. As you are changing APIs I'd like to have a wider review with more of the core Kitura team. We'll look at this next week.

naithar:
@shmuelk no problem, I'll wait for review. :)
There are probably other repos that use `queryParameters`, should I search for them and make a PR that uses this new API or it's better to wait until that PR is merged?

naithar:
@shmuelk any updates on this issue?

shmuelk:
@naithar As you probably know, we've released Kitura 1.0.0. That changes how we change external APIs which your PR does. We need to understand better how to move forward with such changes in general, as we will probably have API breaking changes in the future.

This will probably take a bit longer than we would have liked to get merged.

In any event, does your PR deal with the same query parameter repeated multiple times (i.e. `//host:port/path?a=1&a=2&a=3`) or does it only handle things like  `//host:port/path?a=[1,2,3]` ?

naithar:
@shmuelk currently it handles things like a`host:port/path?a[]=1&a[]=2&a[]=3` as an array. Things like `host:port/path?a=1&a=2&a=3` will result in key `a` having the last value which is `3`. I have a test checking this, but this behavior can be changed.
Implementing a parse for `host:port/path?a=[1,2,3]` should not be a problem if needed.

naithar:
As for changing an API `@availability(*, deprecated)` could be used to mark an old implementation which in case of this PR is `public var queryParameters: [String:String] = [:]` and new implementation could be moved to some other parameter, but i'm not sure this is a good way.

PR #91, opened 2016-08-31
IBM-Swift/Kitura#421 Bubble up errors from HTTPServer

na-gupta:
## Checklist:

### DIFF
--- a/Sources/KituraNet/FastCGI/FastCGIServer.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServer.swift
@@ -57,25 +57,38 @@ public class FastCGIServer {
     /// Listens for connections on a socket
     ///
     /// - Parameter port: port number for new connections (ex. 9000)
+    /// - Parameter errorHandler: optional callback for error handling
     ///
-    public func listen(port: Int) {
-        
+    public func listen(port: Int, errorHandler: ((Swift.Error) -> Void)? = nil) {
         self.port = port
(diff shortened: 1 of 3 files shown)


Sources/KituraNet/FastCGI/FastCGIServer.swift
```diff
@@ -57,25 +57,38 @@ public class FastCGIServer {
...
+            if let callback = errorHandler {
+                callback(error)
+            } else {
+                Log.error("Error creating socket: \(error)")
+            }
         }
-        
+
+        guard let socket = self.listenSocket else {
+            Log.error("Failed to create socket on port \(port)")
```
shmuelk: I don't think there is a need to Log anything here. Either we called the user error handler above, or we logged something already.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -68,28 +68,38 @@ public class HTTPServer {
...
+        } catch {
+            if let callback = errorHandler {
+                callback(error)
+            } else {
+                Log.error("Error creating socket: \(error)")
+            }
+        }
+
+        guard let socket = self.listenSocket else {
+            Log.error("Failed to create socket on port \(port)")
```
shmuelk: I don't think there is a need to Log anything here. Either we called the user error handler above, or we logged something already.

PR #92, opened 2016-09-01
IBM-Swift/Kitura#682 Improve Kitura's performance

shmuelk:
## Description
Improve Kitura's performance by:

  - Using NSData and NSMutableData instead of Data in several internal places
  - Passing UnsafeRawPointer and length pair from the parsing code to the HTTPIncomingMessage class instead of using a temporary Data struct
  - Minimizing the number of times a header key is lower cased when parsing a request

## Motivation and Context
Improving Kitura's performance

## How Has This Been Tested?
Tested using Kitura-net's tests and the TechEmpower tests

## Checklist:

### DIFF
--- a/Sources/KituraNet/BufferList.swift
+++ b/Sources/KituraNet/BufferList.swift
@@ -18,12 +18,18 @@ import Foundation
 
 // MARK: BufferList 
 
+/// This class provides an implementation of a buffer that can be added to and
+/// "taken" from in chunks. Data is always added to the end of the buffer and
+/// data is "taken" from the buffer from the beginning towards the end. A pointer
+/// is maintained as where the next chunk of data will be "taken" from. This
+/// pointer can be reset, to enable the data in the buffer to be fetched from the
+/// beginning again.
 public class BufferList {
 
     // MARK: -- Private 
     
     /// Internal storage buffer
-    private let localData = NSMutableData()
+    private let localData = NSMutableData(capacity: 4096) ?? NSMutableData()
     
     /// Byte offset inside of internal storage buffer
     private var byteIndex = 0
@@ -64,16 +70,12 @@ public class BufferList {
     ///
     /// - Parameter array: a [UInt8] for the data you want from the buffer
     ///
-    /// - Returns:
+    /// - Returns: The number of bytes actually copied from the buffer. It will be the
+    ///           lessor of the number of bytes left in the buffer and the length
+    ///           of the array.
     public func fill(array: inout [UInt8]) -> Int {
         
-        let result = min(array.count, localData.length - byteIndex)
-        let bytes = localData.bytes.assumingMemoryBound(to: UInt8.self) + byteIndex
-        UnsafeMutableRawPointer(mutating: array).copyBytes(from: bytes, count: result)
-        byteIndex += result
-        
-        return result
-        
+        return fill(buffer: UnsafeMutablePointer(mutating: array), length: array.count)
     }
     
     /// Fill memory with data from the buffer
@@ -81,7 +83,9 @@ public class BufferList {
     /// - Parameter buffer: NSMutablePointer to the beginning of the memory to be filled
     /// - Parameter length: The number of bytes to fill
     ///
-    /// - Returns:
+    /// - Returns: The number of bytes actually copied from the buffer. It will be the
+    ///           lessor of the number of bytes left in the buffer and the length
+    ///           of the memory area provided.
     public func fill(buffer: UnsafeMutablePointer<UInt8>, length: Int) -> Int {
         
         let result = min(length, localData.length - byteIndex)
@@ -97,7 +101,8 @@ public class BufferList {
     ///
     /// - Parameter data: The Data struct to fill from data in the buffer
     ///
-    /// - Returns:
+    /// - Returns: The number of bytes actually copied from the buffer. It will be equal
+    ///           to the number of bytes lect in the buffer.
     public func fill(data: inout Data) -> Int {
         
         let result = localData.length - byteIndex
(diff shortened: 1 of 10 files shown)


Sources/KituraNet/BufferList.swift
```diff
@@ -28,120 +22,114 @@ public class BufferList {
 
     // MARK: -- Private 
     
-    ///
     /// Internal storage buffer
-    ///
-    private var localData = Data(capacity: 4096)
+    private let localData = NSMutableData()
```
vadimeisenbergibm: I would use here `localData = NSMutableData(capacity: 4096) ?? NSMutableData()`
So you try to get the buffer, and if you cannot, you fall back to a non-optional initializer or some other default value.

Sources/KituraNet/BufferList.swift
```diff
@@ -28,120 +22,114 @@ public class BufferList {
...
     }
     
+    /// Fill an array with data from the buffer
     ///
-    /// Fill the buffer with a byte array data
-    ///
-    /// - Parameter buffer: a [UInt8] for data you want in the buffer
+    /// - Parameter array: a [UInt8] for the data you want from the buffer
     ///
     /// - Returns:
```
vadimeisenbergibm: Return value documentation is missing. Here it is really needed since it is not trivial what function `fillData` returns.

Sources/KituraNet/BufferList.swift
```diff
@@ -22,138 +22,114 @@ public class BufferList {
...
-    ///
     public func fill(array: inout [UInt8]) -> Int {
```
vadimeisenbergibm: How about reusing `fill(buffer:length:)`:
```swift
public func fill(array: inout [UInt8]) -> Int {
        return fill(buffer: UnsafeMutablePointer(mutating: array), length: array.count)
}
```

Sources/KituraNet/BufferList.swift
```diff
@@ -22,138 +22,114 @@ public class BufferList {
...
-    /// Fill the buffer with a byte array data
-    ///
-    /// - Parameter buffer: NSMutablePointer to the beginning of the array
-    /// - Parameter length: the number of bytes in the array
+    /// - Parameter buffer: NSMutablePointer to the beginning of the memory to be filled
+    /// - Parameter length: The number of bytes to fill
     ///
     /// - Returns:
-    ///
     public func fill(buffer: UnsafeMutablePointer<UInt8>, length: Int) -> Int {
```
vadimeisenbergibm: How about extracting common code into a private function?
```swift
 private func fill(length: Int, doFill: (_: Int) -> ()) -> Int {
        let result = min(length, localData.length - byteIndex)

        doFill(result)

        byteIndex += result
        return result
    }

    ///                                                                                                                                                      
    /// - Returns:                                                                                                                                           
    public func fill(buffer: UnsafeMutablePointer<UInt8>, length: Int) -> Int {
        return fill(length: length) { count in
                   let bytes = localData.bytes.assumingMemoryBound(to: UInt8.self) + byteIndex
                   UnsafeMutableRawPointer(buffer).copyBytes(from: bytes, count: count)
            }
    }

    /// Fill a Data struct with data from the buffer                                                                                                         
    ///                                                                                                                                                      
    /// - Parameter data: The Data struct to fill from data in the buffer                                                                                    
    ///                                                                                                                                                      
    /// - Returns:                                                                                                                                           
    public func fill(data: inout Data) -> Int {
        return fill(length: localData.length - byteIndex) { count in
                   data.append(localData.bytes.assumingMemoryBound(to: UInt8.self) + byteIndex,
                               count: count)
            }
    }

    /// Fill an NSMutableData with data from the buffer                                                                                                      
    ///                                                                                                                                                      
    /// - Parameter data: The NSMutableData object to fill from data in the buffer                                                                           
    ///                                                                                                                                                      
    /// - Returns:                                                                                                                                           
    public func fill(data: NSMutableData) -> Int {
        return fill(length: localData.length - byteIndex) { count in
                   data.append(localData.bytes.assumingMemoryBound(to: UInt8.self) + byteIndex, length: count)
            }
    }
```

vadimeisenbergibm:
@shmuelk After my comments are implemented, the PR looks OK to me.

Sources/KituraNet/BufferList.swift
shmuelk: Closures of that nature that capture fields, should really be avoided in code that is performance important. Every such closure adds an extra malloc and free. We are already spending too much time in malloc / free.

Sources/KituraNet/BufferList.swift
shmuelk: I have made the change you recommended. However, I feel it isn't the correct change to make. If the NSMutableData(capacity: 4096) fails, there is an out of memory situation. It is very likely that allocation the other NSMutableData() will cause a crash.

We need a better solution here, that I don't want to do as part of this PR.

We need a better solution here. Which I don't want to do as part of this PR.

djones6:
Performance results, simple hello-world response:

This gives around 15% improvement on Linux, and around 12% on Mac.   It makes sense that this is of less benefit on Mac as we believe the performance of Data is particularly an issue in the current Linux implementation.

**Linux**
```
Implementation | Avg Throughput | Max Throughput | Avg CPU | Avg RSS (kb) | Avg Lat (ms) | Max Lat (ms)
---------------|----------------|----------------|---------|--------------|--------------|--------------
        0.28.1 |        33755.9 |       34177.67 |    90.4 |        99612 |          3.8 |       201.59
         +pr92 |        38951.9 |       39471.46 |    89.7 |       105081 |          3.3 |        62.17
    0.28.1+gcd |        27883.3 |       28293.10 |    79.5 |        88021 |          4.4 |       205.95
     +pr92+gcd |        31697.9 |       32527.77 |    77.6 |        92874 |          3.9 |        74.87
```
**Mac**
```
Implementation | Avg Throughput | Max Throughput | Avg CPU | Avg RSS (kb) | Avg Lat (ms) | Max Lat (ms)
---------------|----------------|----------------|---------|--------------|--------------|--------------
        0.28.1 |        21362.4 |       21711.90 |    77.8 |       109647 |          6.0 |        64.18
         +pr92 |        23873.0 |       24211.41 |    79.9 |       130738 |          5.3 |        34.97
```

PR #95, opened 2016-09-03
multiple epoll tasks

dgrove-oss:
## Description

distribute incoming socket reads across multiple epoll tasks by hashing
file descriptor to increase concurrency.  

## Motivation and Context
Improves performance, especially as larger systems (higher core counts).

## How Has This Been Tested?
Passes unit tests.
Performance analyzed using Swift-TechEmpower on 12 core linux system shows significant gain from increasing from 1 to 2 tasks.
Some additional (5-10% benefit) to using 4 tasks, but 2 tasks is better than 4 at lower core
counts.

## Checklist:


shmuelk:
I had a version of this locally. I will merge yours and then fix it a bit. You missed the deleteing of the EPOLL events.

PR #99, opened 2016-09-07
Avoid leaking HTTPParser from a ClientRequest error path

djones6:
## Description

This avoids a memory leak where a pair of HTTPIncomingMessage and HTTPParser objects are left with a strong reference cycle when a ClientRequest is unsuccessful (curl returns an error, or unable to parse the response).

## Motivation and Context

Partial fix for Kitura #548 (Memory leak when executing HTTP.request)
- there is still a leak that is somehow related to the headers passed to CCurl, which will be addressed in a future PR.

## How Has This Been Tested?

I ran the Kitura-net tests on Linux and Mac, and a simple testcase for demonstrating the leak.

## Checklist:

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -274,6 +274,9 @@ public class ClientRequest {
     ///
     /// - Parameter close: add the "Connection: close" header
     public func end(close: Bool = false) {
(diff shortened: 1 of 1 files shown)


shmuelk:
After reviewing this with David, we've decided to take a different approach

PR #101, opened 2016-09-08
Fixed memory leak in ClientResponse in error situations.

shmuelk:
Fixed a memory leak in ClientResponse by making sure the HTTPParser object was freed in various error situations. Extra error checking was added.

## Description
Deferred creation of ClientResponse object until the end() function of ClientRequest was called. Insured that ClientResponse.release() was called in various error conditions in ClientRequest.end().
Added error checking in ClientRequest.end() that at least the headers were parsed, before returning as success.

## Motivation and Context
Fix some of the problems in Kitura issue 548

## Checklist:

… more error checking.

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -63,7 +63,7 @@ public class ClientRequest {
     fileprivate var writeBuffers = BufferList()
 
     /// Response instance for communicating with client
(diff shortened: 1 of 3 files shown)


djones6:
Looks good.  We discussed reduction of the number of places where we have to remember to release() the ClientResponse, to avoid reintroducing this bug in the future, however that's out of the scope for this PR.

PR #109, opened 2016-09-13
Combine dictionaries in HeadersContainer class

tardieu:
## Description

Combine the header dictionary and case insensitive map into a single dictionary.

This requires a small change to Kitura (Pull request 737).

## Motivation and Context

Improve performance by reducing dictionary lookups and updates.

## How Has This Been Tested?

Ran swift test

## Checklist:

### DIFF
--- a/Sources/KituraNet/FastCGI/FastCGIServerResponse.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServerResponse.swift
@@ -130,7 +130,7 @@ public class FastCGIServerResponse : ServerResponse {
 
         // add the rest of our response headers
         for (key, valueSet) in headers.headers {
-            for value in valueSet {
+            for value in valueSet.value {
                 headerData.append(key)
                 headerData.append(": ")
                 headerData.append(value)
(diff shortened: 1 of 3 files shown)


tardieu:
The HeadersContainer class maintains a dictionnary "headers" from header keys (String) to header values ([String]). In addition the dictionnary "caseInsensitiveMap" maps the lowercased header keys to their preferred case. Keys are compared using their lowercased form, and values for keys with the same lowercased form are combined. The preferred case is the case of the first occurrence of a key.

The proposed patch intends to maintain the same behavior using a single dictionnary. The new unique dictionnary maps the lowercased keys to a tuple containing the key with its preferred case and the values.  A single lookup (for the lowercased key) replace two lookups (lowercased key in the caseInsensitiveMap followed by the cased key in the headers dictionnary). As before, keys with equal lowercased forms are combined and the preferred key case is retained for pretty printing the headers.

Because the "headers" dictionnary type is modified from [String : [String]] to [String: (key: String, value: [String])] the type of the dictionnary index is affected. This type is currently exposed by Kitura so it must be updated in Kitura in addition to Kitura-net.

PR #135, opened 2016-11-01
Send Upgrade and Connection headers in the switching protocols response

shmuelk:
## Description
Correct the response sent when upgrading protocols. The response is suppose to contain Upgrade and Connection headers. It didn't before this fix.

## Motivation and Context
Get real protocol upgrades to work.

## How Has This Been Tested?
This has been tested as part of the Kitura-WebSocket development work. Without this fix, the Node WebSocket client couldn't upgrade the connection from HTTP to websocket. With it it can. Checking a source or two indicates that indeed the response was suppose to have the headers.

## Checklist:

### DIFF
--- a/Sources/KituraNet/ConnectionUpgrader.swift
+++ b/Sources/KituraNet/ConnectionUpgrader.swift
@@ -62,10 +62,12 @@ public struct ConnectionUpgrader {
         var responseBody: String?
         var notFound = true
         let protocolList = protocols.split(separator: ",")
+        var foundProtocol: String?
         for eachProtocol in protocolList {
             let theProtocol = eachProtocol.first?.trimmingCharacters(in: CharacterSet.whitespaces) ?? ""
             if theProtocol.characters.count != 0, let factory = registry[theProtocol.lowercased()] {
                 (processor, responseBody) = factory.upgrade(handler: handler, request: request, response: response)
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/ConnectionUpgrader.swift
```diff
@@ -62,10 +62,12 @@ public struct ConnectionUpgrader {
         var responseBody: String?
         var notFound = true
         let protocolList = protocols.split(separator: ",")
+        var foundProtocol: String?
```
ianpartridge: `foundProtocol` reads as a boolean to me: did we find the protocol or not? If this is the protocol name, call it `protocolName`?

Sources/KituraNet/ConnectionUpgrader.swift
```diff
@@ -82,6 +84,8 @@ public struct ConnectionUpgrader {
             else {
                 if let theProcessor = processor {
                     response.statusCode = .switchingProtocols
+                    response.headers["Upgrade"] = [foundProtocol!]
```
ianpartridge: Could we ever get here with `foundProtocol == nil`? Force unwraps are generally a bad idea, I think convention is to do something like:

```swift
if let theProcessor = processor, let protocolName = protocolName
```

PR #140, opened 2016-11-08
Add initial set of Monitoring hooks to KituraNet.

shmuelk:
## Description
Add an initial set of Monitoring hooks to KituraNet that can be exploited by Monitoring packages that can be added to a Kitura based application. We've started with two events:
   1. A HTTP request has finished parsing and is about to be handled
   2. Processing of a HTTP request has finished (end was called)

## Motivation and Context
Enable application server monitoring.

## How Has This Been Tested?
KituraNet unit tests extended and run on both Linux and macOS.

## Checklist:


Sources/KituraNet/Server/ServerMonitor.swift
```diff
@@ -0,0 +1,38 @@
...
+/// A protocol for implementing a delegate to receive monitoring events from KituraNet.
+public protocol ServerMonitor {
+    /// An event fired when a HTTP request has finished being parsed and is about
+    /// to be passed to the `ServerDelegate` for processing.
+    ///
+    /// - Parameter request: The `ServerRequest` class instance for the request starting.
+    /// - Parameter response: The `ServerResponse` class instance for the request starting.
+    func started(request: ServerRequest, response: ServerResponse)
+    
+    /// An event fired when th processing of a HTTP request has finished.
```
youming-lin: `when th processing` -> `when the processing`

Tests/KituraNetTests/MonitoringTests.swift
```diff
@@ -0,0 +1,114 @@
...
+        func handle(request: ServerRequest, response: ServerResponse) {
+            let payloadData = contentTypesString.data(using: .utf8)!
+            do {
+                response.statusCode = .OK
+                response.headers["Content-Length"] = ["\(payloadData.count)"]
+                try response.write(from: payloadData)
+                try response.end()
+            }
+            catch {
+                print("Error writing response.")
```
youming-lin: Wouldn't it be better to use `XCTFail("Error writing response.")` here instead?

shmuelk: Correct and fix pushed.

youming-lin:
Fixes look good :+1:

PR #144, opened 2016-11-17
Fix the build break in KituraNet

shmuelk:
Don't unconditionally unwrap a weak variable.

## Motivation and Context
Fix a build break

## How Has This Been Tested?
Ran KituraNet unit tests

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
+++ b/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
@@ -156,8 +156,11 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
         state = .messageCompletelyRead
         response.reset()
         
+        // If the IncomingSocketHandler was freed, we can't handle the request
+        guard let handler = handler else { return }
+        
         if isUpgrade {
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
```diff
@@ -156,8 +156,11 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
         state = .messageCompletelyRead
         response.reset()
         
+        // If the IncomingSocketHandler was freed, we can't handle the request
+        guard let handler = handler else { return }
```
ianpartridge: Is it OK for this to be outside the `if` block? Could we end up returning in the non-`isUpgrade` case, and hence not calling `self.delegate?.handle()`?

shmuelk: handler is the IncomingSocketHandler for the socket in question. If the handler is now nil, we have no way to send back any kind of response.

That's why I just "dropped" the new request in such a situation.

ianpartridge: OK, thanks, that makes sense.

PR #148, opened 2016-11-28
Reduce priority of `Connection reset by peer` messages.

carlbrown:
Under common operation, there are a lot of `Connection reset by peer` messages that get logged by Kitura-Net, especially when put behind an HAProxy (see http://discourse.haproxy.org/t/haproxy-1-5-4-1-5-18-is-sending-rst-ack-in-for-closure-to-httpchk-instead-of-fin-ack/474/1 for more information).

This change just reduces the priority at which these specific messages are logged so they don't 1) swamp the client applications' logs, and 2) make the Application Developer think something is wrong when something expected happens.

### DIFF
--- a/Sources/KituraNet/IncomingSocketHandler.swift
+++ b/Sources/KituraNet/IncomingSocketHandler.swift
@@ -106,7 +106,7 @@ public class IncomingSocketHandler {
             }
         }
         catch let error as Socket.Error {
-            if error.errorCode == Int32(Socket.SOCKET_ERR_RECV_FAILED) {
+            if error.errorCode == Int32(Socket.SOCKET_ERR_CONNECTION_RESET) {
                 Log.debug("Read from socket (file descriptor \(socket.socketfd)) reset. Error = \(error).")
             } else {
                 Log.error("Read from socket (file descriptor \(socket.socketfd)) failed. Error = \(error).")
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/IncomingSocketHandler.swift
```diff
@@ -106,7 +106,11 @@ public class IncomingSocketHandler {
             }
         }
         catch let error as Socket.Error {
-            Log.error("Read from socket (file descriptor \(socket.socketfd)) failed. Error = \(error).")
+            if error.errorCode == Int32(Socket.SOCKET_ERR_RECV_FAILED) {
```
shmuelk: Did @billabt say that the error code of SOCKET_ERR_RECV_FAILED is what you get when the connection was reset by the peer? It seems kind of overly general to me.

billabt:
There was no error that *explicitly* means connection reset by peer.  The *SOCKET_ERR_RECV_FAILED* could be for any number of reasons.  So, I just added *SOCKET_ERR_CONNECTION_RESET* which will be the error when the connection is actually reset by the peer.  This is detected by checking *errno* **immediately** following the function call for **ECONNRESET** and then throwing that error instead of other one.

carlbrown:
I'm fine with rejecting this one and I'll resubmit a new one against the new version of Kitura-Net once Bill's change is in and Kitura is tagged having included it.

Until then, I'll keep running this hack on the existing sandbox servers.

I created this PR to try to facilitate communication and transparency, not because I was confident it was the perfect solution.

billabt:
The changes to BlueSocket are checked in.  One note, the error can occur on a write to the socket as well as read so you may want to modify your PR to take this into account.

PR #149, opened 2016-11-29
IBM-Swift/Kitura#869 Set socket processor handler before starting and…

na-gupta:
## Description
- Set socket processor handler before starting
- Log error in processor when handler is nil
- Add call site line number to XCTestExpectation description
- Change osx image for travis to 8.1

## Motivation and Context
Kitura osx builds are hanging intermittently when IncomingHTTPSocketProcessor finishes parsing before the handler was set and was failing silently. See IBM-Swift/Kitura#869

## Checklist:

### DIFF
--- a/Tests/KituraNetTests/KituraNetTest.swift
+++ b/Tests/KituraNetTests/KituraNetTest.swift
@@ -23,7 +23,7 @@ import Dispatch
 
 protocol KituraNetTest {
 
-    func expectation(_ index: Int) -> XCTestExpectation
+    func expectation(line: Int, index: Int) -> XCTestExpectation
     func waitExpectation(timeout t: TimeInterval, handler: XCWaitCompletionHandler?)
 }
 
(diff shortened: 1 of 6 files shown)


Tests/KituraNetTests/KituraNetTest.swift
```diff
@@ -37,41 +37,41 @@ extension KituraNetTest {
...
+        do {
             try server.listen(on: 8090)
 
             waitExpectation(timeout: 10) { error in
                 // blocks test until request completes
                 server.stop()
                 XCTAssertNil(error);
             }
-        } catch {
+        } catch let error {
```
ianpartridge: Just FYI, I think this is unnecessary:

From [Error Handling in Swift](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/ErrorHandling.html):
> If a catch clause doesn’t have a pattern, the clause matches any error and binds the error to a local constant named error.

PR #150, opened 2016-11-29
IBM-Swift/Kitura#876 Log client and proxy info using HTTP and FastCGI…

na-gupta:
## Description
- Log client(forwarded for) and proxy info using HTTP and FastCGI headers
- Lower priority of socket level connection logging

## Motivation and Context
See IBM-Swift/Kitura#876

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
+++ b/Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
@@ -44,10 +44,10 @@ public class HTTPIncomingMessage : HTTPParserDelegate {
     public let isRequest: Bool
 
     /// Client connection socket
-    private var socket: Socket?
+    private let socket: Socket?
 
     /// socket signature of the request.
-    public private(set) var signature: Socket.Signature?
+    public var signature: Socket.Signature? { return socket?.signature }
 
     /// The URL from the request as URLComponents.
     public private(set) var urlComponents = URLComponents()
@@ -105,12 +105,9 @@ public class HTTPIncomingMessage : HTTPParserDelegate {
     /// - Parameter isRequest: whether this message is a request
     ///
     /// - Returns: an IncomingMessage instance
-    init (isRequest: Bool, socket: Socket? = nil, signature: Socket.Signature? = nil) {
+    init (isRequest: Bool, socket: Socket? = nil) {
         self.isRequest = isRequest
-        if let socket = socket {
-            self.socket = socket
-        }
-        self.signature = signature
+        self.socket = socket
         httpParser = HTTPParser(isRequest: isRequest)
 
         httpParser!.delegate = self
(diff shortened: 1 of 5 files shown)


Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
```diff
@@ -281,8 +287,10 @@ public class HTTPIncomingMessage : HTTPParserDelegate {
 
         if isRequest {
             var url = ""
+            var proto: String?
```
ianpartridge: This code allows `proto` to hold 3 possible states: nil, "https" or "http". Is that wise? Can we make this a non-optional String that either holds "https" or "http"?

na-gupta: My intent was that signature.isSecure should always be available for a request, but if there was a bug that caused it not to be, and so we didn't know what the scheme was, we should flag it as such so it might be noticed instead of putting in a default value that may or may not be accurate.

ianpartridge: Makes sense - thanks.

Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
```diff
@@ -102,8 +105,11 @@ public class HTTPIncomingMessage : HTTPParserDelegate {
     /// - Parameter isRequest: whether this message is a request
     ///
     /// - Returns: an IncomingMessage instance
-    init (isRequest: Bool, signature: Socket.Signature? = nil) {
+    init (isRequest: Bool, socket: Socket? = nil, signature: Socket.Signature? = nil) {
         self.isRequest = isRequest
+        if let socket = socket {
+            self.socket = socket
+        }
```
shmuelk: Also why pass in both the socket and the socket's signature?

Pass in just the socket.

Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
billabt: Would it help if I put a helper computed property in **BlueSocket** to provide `isSecure`?  This way you wouldn't have to worry about accessing the `Signature`.

Come to think of it, I may do it anyway...  Seems cleaner.

Sources/KituraNet/HTTP/HTTPIncomingMessage.swift
na-gupta: Yes, I can change it to just pass in the socket and not the signature.

The passed in socket in init is optional because currently HTTPServerResponse does not have a reference to the socket. We could change that, but HTTPServerResponse does not have a need for the socket as far as I can see.

shmuelk: I understand why self.socket is optional. In any event you are correct that ClientResponse, the other inheritor of HTTPIncomingMessage, doesn't have a socket.


The question is about the if let..... Why not just simply assign it.......

na-gupta: Ah, yes, good point :)  Will do that

shmuelk:
Looks great now.

Waiting for Travis-CI to complete to merge.

PR #154, opened 2016-12-05
Added basic E2E test for FastCGI and fixed minor typos in FastCGIServer.swift

kweinmeister:
## Description
Added a test that starts and stops a server, and checks that the states are as expected.

## Motivation and Context
There is currently a much lower test coverage % in FastCGI than in other parts of Kitura-Net. The existing FastCGI tests focus on the FastCGI protocols.

## How Has This Been Tested?
A new test has been created, and it is successful

### DIFF
--- a/Tests/KituraNetTests/ClientE2ETests.swift
+++ b/Tests/KituraNetTests/ClientE2ETests.swift
@@ -27,8 +27,7 @@ class ClientE2ETests: XCTestCase {
         return [
             ("testSimpleHTTPClient", testSimpleHTTPClient),
             ("testPostRequests", testPostRequests),
-            ("testErrorRequests", testErrorRequests),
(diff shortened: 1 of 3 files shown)


Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -27,7 +27,8 @@ class ClientE2ETests: XCTestCase {
         return [
             ("testSimpleHTTPClient", testSimpleHTTPClient),
             ("testPostRequests", testPostRequests),
-            ("testErrorRequests", testErrorRequests)
+            ("testErrorRequests", testErrorRequests),
+            ("testFastCGIStates", testFastCGIStates)
```
shmuelk: This test should be in LifecycleListenerTests.swift and not in ClientE2ETests.

PR #155, opened 2016-12-07
Fix a memory leak that occurs with keep alive

shmuelk:
Fix a memory leak that occurs when sockets kept open due to keep alive are finally closed due to inactivity.

## Description
When sockets are closed due to inactivity make sure that the IncomingSocketProcessor.prepareToClose() function is called. This will break the retain loop between the HTTPIncomingMesage and the HTTPParser.

## Motivation and Context
Fixes a memory lak reported in isuue IBM-Swift/Kitura#902

## How Has This Been Tested?
@djones tested using Safari and determined that the memory leak was fixed.
Ran KituraNet unit tests on macOS and Linux.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -118,8 +118,8 @@ public class HTTPServerResponse : ServerResponse {
             if !keepAlive && !processor.isUpgrade {
                 processor.close()
             }
+            Monitor.delegate?.finished(request: processor.request, response: self)
         }
-        Monitor.delegate?.finished(request: processor?.request, response: self)
     }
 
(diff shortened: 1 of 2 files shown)


djones6:
My method for testing this was to refresh the project that @carlbrown provided for https://github.com/IBM-Swift/Kitura/issues/902 on Mac using Safari (which seems, at least for me, to sometimes ditch existing connections and open new ones).  The leak _seems_ to be gone, though my method of testing wasn't very scientific.  I no longer see an accumulation of instances of ServerRequest / HTTPParser in the xcode debugger.

I'll work on a more repeatable benchmark as we need regression testing for this.  This case is non-trivial as it needs to open connections, drive some requests, then let them idle so that we can witness the memory usage reduce once another incoming connection kicks off our idleness check.

PR #160, opened 2016-12-09
Decouple our HTTPParserDelegate from our http_parser C library

carlbrown:
## Description
This doesn't actually change any behavior external to this class. It's just cleaner, easier to follow (i.e. it has comments and references for the pointer manipulations) and it enables future work to allow the HTTPParserDelegate to become `weak` which will allow this parser to be reclaimed by ARC

## Motivation and Context
Precondition to clean up memory leaks from #879 and #902

## How Has This Been Tested?
ran `swift test` (also ran tests in Xcode with the AddressSanitizer on)
Put it into `Sandbox-UI` and ran my load-test suite against it (locally)

### DIFF
--- a/Sources/KituraNet/HTTPParser/HTTPParser.swift
+++ b/Sources/KituraNet/HTTPParser/HTTPParser.swift
@@ -31,21 +31,13 @@ class HTTPParser {
     var isRequest = true
 
     /// Delegate used for the parsing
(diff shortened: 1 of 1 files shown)


shmuelk:
Resolved via issue IBM-Swift/Kitura#933

PR #162, opened 2016-12-23
Compute urlURL and logging information when needed

shmuelk:
Defer creation of urlURL until it is referenced and compute data needed for verbose logging only when logging at the verbose level

## Motivation and Context
It turns out that creating a URL struct is actually quite expensive. Defer the creation of urlURL until it is actually used by the user of KituraNet. Additionally information needed for logging at the verbose level should only be computed if indeed logging at the verbose level.

## How Has This Been Tested?
Ran KituraNet unit tests. Was run under TechEmpower tests as well.

## Checklist:


ianpartridge:
It would be nice to have a test for `urlURL` but I'll merge this anyway. Thanks.

PR #170, opened 2017-01-19
IBM-Swift/Kitura#976 Another attempt to fix socket I/O guards

na-gupta:
## Description
Another attempt to fix socket I/O guards, hopefully this covers all the edge cases

## Motivation and Context
IBM-Swift/Kitura#976

## Checklist:

### DIFF
--- a/Sources/KituraNet/IncomingSocketHandler.swift
+++ b/Sources/KituraNet/IncomingSocketHandler.swift
@@ -315,6 +315,10 @@ public class IncomingSocketHandler {
                 Log.error("Write to socket (file descriptor \(socket.socketfd)) failed. Error = \(error).")
             }
         }
+
+        if preparingToClose {
+            close()
+        }
     }
 
     /// If there is data waiting to be written, set a flag and the socket will
--- a/Sources/KituraNet/IncomingSocketHandler.swift
+++ b/Sources/KituraNet/IncomingSocketHandler.swift
@@ -63,7 +63,8 @@ public class IncomingSocketHandler {
     private var preparingToClose = false
     private var isOpen = true
     private var writeInProgress = false
-    private var readInProgress = false
+    private var handleWriteInProgress = false
+    private var handleReadInProgress = false
 
     /// The file descriptor of the incoming socket
     var fileDescriptor: Int32 { return socket.socketfd }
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/IncomingSocketHandler.swift
```diff
@@ -60,10 +60,24 @@ public class IncomingSocketHandler {
...
+    /// isOpen is set to false when:
+    ///   - close() is invoked AND
+    ///   - it is safe to close the socket (there is no data waiting to be written and a socket read/write is not in progress).
+    /// This lets other threads know to not start reads/writes on this socket anymore, which could cause a crash.
     private var isOpen = true
+
+    /// writeInProgress, handleWriteInProgress and handleReadInProgress are:
+    ///   - set when the corresponding functions start their work.
+    ///   - unset when the corresponding functions finish their work.
+    /// This lets other threads know to not try to close this socket while those functions are using it, which could cause a crash.
```
shmuelk: The comment above seems to be an API comment (i.e. ///) if so, it should be repeated per variable involved.

Sources/KituraNet/IncomingSocketHandler.swift
```diff
@@ -290,50 +334,58 @@ public class IncomingSocketHandler {
...
+            // a read/write function in between us clearing the guard and setting the flag.
+            // Make sure to set it back to open if the guard fails and we don't actually close.
+            guard !writeInProgress && !handleWriteInProgress && !handleReadInProgress
+                && writeBuffer.length == writeBufferPosition else {
+                    isOpen = true
+                    return
+            }
+
+            #if os(OSX) || os(iOS) || os(tvOS) || os(watchOS) || GCD_ASYNCH
+                readerSource.cancel()
```
shmuelk: Did you verify what happens if you call readerSource.cancel() multiple times? That is, is handleCancel() called each time readerSource.cancel() is called?

na-gupta: The handler is called just the first time like you had thought. That's why I moved the guards from the handleRead() to close()

shmuelk: The question is about **handleCancel()**, because handleCancel() closes the socket.

shmuelk: Never mind. I now understand what you meant.

na-gupta: Yes, sorry, I meant handleCancel, not handleRead. My fingers and brain don't communicate well before my morning coffee.

Sources/KituraNet/IncomingSocketHandler.swift
na-gupta: Fixed so we only use API style comments for public/open members.

PR #174, opened 2017-01-31
Fix the frequently broken KituraNet build

shmuelk:
## Description
Fixed the failing LifecycleListener failed callback test  by changing it to not trying to listen on port -1, but rather to try start a second server on the same address.

## Motivation and Context
Fixing the KituraNet build that was failing some of the time.

## How Has This Been Tested?

Ran the KituraNet tests over thirty time in a row on both Linux and macOS without any failures.

## Checklist:


na-gupta:
The builds do not seem to be failing anymore. But the updated test is better in any case, so merging it.

PR #176, opened 2017-02-16
Merge the memory leak fixes into master

shmuelk:
## Description
Remove reuse of ServerRequest and ServerResponse objects in KituraNet in cases of KeepAlive. Move up in the stack the HTTP Parsing and as a result remove the super class HTTPIncomingMessage, that was common to ClientResponse and HTTPServerRequest.

## Motivation and Context
This fixes major memory leak issues.

This code needs some cleaning up. But it solves such a real problem, that I think it should be merged now.

## How Has This Been Tested?
This has been tested by the KituraNet unit tests and the performance benchmark.

## Checklist:


Sources/KituraNet/ClientResponse.swift
```diff
@@ -16,19 +16,149 @@
 
 
 import Foundation
+import LoggerAPI
 
 // MARK: ClientResponse
 
 /// This class describes the response sent by the remote server to an HTTP request
 /// sent using the `ClientRequest` class. This class is an extension of the 
 /// `HTTPIncomingMessage` class.
```
na-gupta: API comments need updating here and in other places.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -15,30 +15,208 @@
...
-/// and the `ClientResponse` class.
-public class HTTPServerRequest: HTTPIncomingMessage, ServerRequest {
-
-    /// Client socket
-    private let clientSocket: Socket
+/// are communicating via the HTTP protocol. 
+public class HTTPServerRequest: ServerRequest {
+    
+    /// HTTP Status code if this message is a response
+    public private(set) var httpStatusCode: HTTPStatusCode = .unknown
```
na-gupta: Should we deprecate the members that were intended for ClientResponse?

Sources/KituraNet/ClientResponse.swift
```diff
@@ -16,19 +16,149 @@
...
-public class ClientResponse: HTTPIncomingMessage {
+public class ClientResponse {
     
-    /// Initializes a `ClientResponse` instance
-    init() {
+    /// HTTP Status code if this message is a response
+    public private(set) var httpStatusCode: HTTPStatusCode = .unknown
+    
+    /// HTTP Method of the incoming message.
+    public var method: String { return httpParser.method }
```
na-gupta: Should we deprecate the members that were intended for HTTPServerRequest?

Sources/KituraNet/ClientResponse.swift
youming-lin: Need to regenerate docs via jazzy as well

Sources/KituraNet/ClientResponse.swift
shmuelk: Probably

Sources/KituraNet/HTTP/HTTPServerRequest.swift
shmuelk: Done

Sources/KituraNet/ClientResponse.swift
shmuelk: We should regenerate the the docs right before we merge this PR.

shmuelk: Removed the reference to the super class.

djones6:
It turns out that the performance test we ran for this issue was flawed;  due to an error in the git commands, while building, the issue_1012 branch was reset back to the point it was branched (we effectively compared Kitura-net 1.6.1 to itself).

There is a performance penalty to this change of up to 20% (Hello World, which is the worst case scenario).

Ubuntu 16.04:
```
    -net 1.6.1 |    83123.9 |    84042.8 |    98.7 |        25398 |      1.6 |      7.9 |     33.5 |     3 
    -net 1.6.2 |    66985.0 |    68645.2 |    95.7 |        24392 |      1.9 |      4.1 |     22.7 |     3 
```
Macbook:
```
    -net 1.6.1 |    39925.3 |    41563.7 |    80.5 |        64074 |      6.7 |    223.8 |    546.4 |     3 
    -net 1.6.2 |    37547.3 |    38003.8 |    78.6 |       114065 |     10.0 |    410.7 |    730.7 |     3 
```

PR #178, opened 2017-03-10
IBM-Swift/Kitura#991 Trap SIGPIPE signal

djones6:
## Description
Prevent unexpected process termination when the receiver closes their socket while we were trying to write.  Resolves IBM-Swift/Kitura#991

## Motivation and Context
See IBM-Swift/BlueSSLService/issues/13 and IBM-Swift/BlueSSLService/pull/14
Bill and I decided this fix was an application decision and belonged in Kitura, not SSLService.

## How Has This Been Tested?
Ran a perf test which exited with SIGPIPE frequently before when the clients disconnect.  It now logs a message at .info and continues execution.

## Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -33,4 +33,6 @@ let package = Package(
 #if os(Linux)
     package.dependencies.append(
         .Package(url: "https://github.com/IBM-Swift/CEpoll.git", majorVersion: 0, minor: 1))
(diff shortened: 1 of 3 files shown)


djones6:
Good point, I've added the same for FastCGIServer.

PR #183, opened 2017-04-03
Change forced server startup failure to use an invalid port number

shmuelk:
## Description
A change in BlueSocket enabled listening on the same port in multiple processes/threads.

This has broken one of Kitura-net's LifecycleListenerTests where it was trying to force a server startup failure by listening on the same port in two different threads.

The fix in this issue will instead simply listen on an invalid port (-1).

## Motivation and Context
Fix the broken build

## How Has This Been Tested?
Ran Kitura-net unit tests which were failing before the fix.

## Checklist:


irar2:
LGTM

PR #199, opened 2017-06-12
Swift 4 support

quanvo87:
## Checklist:


Sources/KituraNet/FastCGI/FastCGIServerResponse.swift
```diff
@@ -19,32 +19,47 @@ import Socket
...
-    func iterate(by delta: Bound, action: (Range<Bound>) throws -> ()) throws {
-        
-        var base = self.lowerBound
-        
-        while (base < self.upperBound) {
-            let subRange = (base ..< (base + delta)).clamped(to: self)
-            try action(subRange)
-            
-            base += delta
+#if swift(>=4)
```
quanvo87: @kweinmeister I tried just putting the needed lines in the `#if`s, but this is one of those cases where the compiler gets confused.

PR #210, opened 2017-07-24
Replace explicit unwrapping with conditional unwrapping

shmuelk:
## Description
There is code in IncomingSocketManager that tests if a field is NOT nil and then proceeds to explicitly unwrap it. This turns out to be problematic as the field could be modified in another thread at the same time.

The code was changed to conditionally unwrap the field in question.

## Motivation and Context
Fixes issue IBM-Swift/Kitura-net#209 

## How Has This Been Tested?
The Kitura-net unit tests have been run. Although admittedly this is an intermittent failure.

## Checklist:


djones6:
This looks good to me.

PR #215, opened 2017-08-04
Add support for SSL certificate and private key in ClientRequest

emiliopavia:
## Description
Updates to the ClientRequest so that it can make a connection providing a SSL certificate and a private key.

## Motivation and Context
Some servers require a SSL certificate to connect (e.g. Apple APNs). This modification allows you to pass both certificate and private key path (and the passphrase if needed) to libcurl.

## How Has This Been Tested?
This modification was tested by creating a terminal client on Linux that sends a push notification to an iOS sample app. The notification was successfully sent and received by the physical device.

## Checklist:


geertberkers:
Any updates on this? I would like to see this in the library! Currently using NSUrlSession with a special implemented delegate, but that is not working on Linux.

ianpartridge:
Hi - I understand the use-case but I think there are a few issues with this PR as it stands:

1. It adds new cases to `ClientRequest.Options` - unfortunately because Swift does not yet have "open" enums this is a breaking change according to SemVer.
2. It is PEM-only. On macOS we would want to support "P12" for PKCS#12 certificates too - curl does support this - https://curl.haxx.se/libcurl/c/CURLOPT_SSLCERTTYPE.html
3. It is bespoke and not built on https://github.com/IBM-Swift/BlueSSLService - we use `BlueSSLService.Configuration` for server-side certificate handling and it would be best to reuse the same APIs for client-side certificates too.

So as it stands we won't be able to merge this unfortunately. Having said that I do see the value in the feature so we have no objection in principle to supporting client-side certificates in `ClientRequest`.

emiliopavia:
I can try to change the implementation with all your requirements and suggestions. Thank you!

geertberkers:
@ianpartridge Thank you for the response, i was thinking about my own implementation (based on @emiliopavia answer). 

1. Is it okay to add another enum (e.g. ClientRequest.SSLConfig) and add another initialiser with ClientRequest.Options and ClientRequest.SSLConfig? Or if we want to implement BlueSSLService, just add SSLService as object? 
2. For this we could add a simple string/enum for PEM, DER and P12 and set these ourself, or based on selected files.
3. We could use the parameters like certificateFilePath, keyFilePath, password used in the BlueSSLService? (But is this necessary when we only need the paths and password? Looks a overkill for me.

ianpartridge:
Yes I think the best thing is to add `ClientRequest.SSLConfig`. If possible we'd like to reuse as much of BlueSSLService as possible, hopefully that won't be overkill.

@pushkarnk can you give your thoughts on this as well please?

PR #216, opened 2017-09-08
Allow max Keep-Alive requests to be configurable

djones6:
## Description

Enables the previously hard-coded maximum of 100 requests per Keep-Alive connection to be configured.  A negative value disables the feature, and is the default (in keeping with other web frameworks).  A value of zero disables Keep-Alive.

This is part of the solution to IBM-Swift/Kitura#1125 and a prerequisite for a corresponding change in Kitura.

## Motivation and Context

Kitura's performance in some benchmarking scenarios suffers because of the cost of recreating new connections after every 100 requests.  This was particularly problematic for [which_is_the_fastest](https://github.com/tbrand/which_is_the_fastest), where the client slows down substantially when forced to open new connections.

## How Has This Been Tested?

I ran the Kitura-net tests and they passed.
I also ran experiments with a simple Hello World Kitura app, varying the number of Keep-Alive requests allowed, and using `curl -i` to confirm that the headers and connection behaviour was as expected.

## Checklist:


Sources/KituraNet/HTTP/KeepAliveState.swift
```diff
@@ -0,0 +1,52 @@
...
+    /// Decrements the number of requests remaining
+    mutating func decrement() -> Void {
+        switch self {
+        case .limited(let limit):
+            if limit > 1 {
+                self = .limited(maxRequests: limit - 1)
+            } else {
+                self = .disabled
+            }
+        default: break
```
ianpartridge: I think this deserves an `assertionFailure()`.

djones6: I agree, I've pushed a change to add assertions in this area.

Sources/KituraNet/HTTP/KeepAliveState.swift
```diff
@@ -0,0 +1,55 @@
...
+        case .limited(let limit): return limit > 0
+        }
+    }
+    
+    /// Decrements the number of requests remaining
+    mutating func decrement() -> Void {
+        switch self {
+        case .unlimited: break
+        case .limited(let limit):
+            if limit > 1 {
```
shmuelk: Shouldn't this be `if limit > 0` ?

djones6: The way I coded this was to have the final `decrement()` set it to `.disabled` rather than `.limited(0)` (which happens in the `else` clause below).  But perhaps that's unnecessarily complicated.  The `keepAlive()` function gives the correct answer either way.

PR #218, opened 2017-09-12
Move the SSL ALPN to a separate thread

shmuelk:
## Description
Don't run the Application Level Protocol Negotiation in the thread that accepted the new socket. Instead run it in a different thread.

## Motivation and Context
Prevent lock ups when there are problems with the SSL negotiation. Fixes Issue IBM-Swift/Kitura#1143

## How Has This Been Tested?
Have run the Kitura unit tests and load tested by @djones6.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -188,22 +188,11 @@ public class HTTPServer: Server {
                 let clientSocket = try listenSocket.acceptClientConnection()
                 Log.debug("Accepted HTTP connection from: " +
                     "\(clientSocket.remoteHostname):\(clientSocket.remotePort)")
-				
-                #if os(Linux)
-                    let negotiatedProtocol = clientSocket.delegate?.negotiatedAlpnProtocol ?? "http/1.1"
-                #else
-                    let negotiatedProtocol = "http/1.1"
-                #endif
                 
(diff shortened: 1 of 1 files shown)


djones6:
This didn't cause performance regression, but also appears not to have fixed the hang,  which I believe is happening at the `acceptClientConnection()` stage (which happens before the `handleNewClient()` part)

The code flow seems to be:
HTTPServer calls `listenSocket.acceptClientConnection()` (Socket)
- calls `self.delegate?.onAccept(socket: newSocket)` (SSLService)
- calls `SSL_accept(sslConnect)` which then blocks

I wonder if the fix belongs in Socket or SSLService? The issue seems to be that, if `Socket` has a `delegate` and calls `onAccept()`, the delegate is probably going to try to read some more bytes off the wire, which we now know could block.

A fix in either place would add a dependency on Dispatch.  Either that, or somehow decoupling the TCP-level socket accept from the protocol-level handshaking done in onAccept so that the application (Kitura) can decide to run it on a separate thread.

djones6:
@shmuelk I believe I have implemented this change in #219:  since ALPN requires SSL, the changes from #219 will cause the rest of the socket initialization / handling to be performed on another thread.

`SocketManager.handle()` will still be called on the main listener thread if there is no delegate, but I think it's safe to assume that this case can only be vanilla http/1.1.

ianpartridge:
@djones6 can we close this?

djones6:
I think so as per my comments above. We can always raise it again if necessary.

PR #219, opened 2017-09-14
IBM-Swift/Kitura#1143 Invoke socket delegate onAccept on a separate thread

djones6:
## Description

This PR decouples the invocation of `delegate?.onAccept()` from `listenerSocket.acceptClientConnection()`, invoking the former in a separate thread.

Note, this depends on IBM-Swift/BlueSocket#85 and the subsequent related changes, tagged in BlueSocket `0.12.67`.

## Motivation and Context

Resolves IBM-Swift/Kitura#1143

This avoids blocking the listener thread if a badly-behaved client connects to an SSL-enabled listener, which will block in `SSL_accept` until the client transmits an SSL handshake.

## How Has This Been Tested?

I ran the Kitura-net tests (with Socket branch issue_1143) and they passed.

I tested with a Kitura-based application which uses SSLService, driven under load (50 concurrent clients from wrk) and the application functioned correctly. I tested on both Linux and Mac.

I also verified that, when creating a bad connection (as described above), the listener was no longer prevented from accepting further well-behaved connections.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -192,12 +192,16 @@ public class HTTPServer: Server {
                 Log.debug("Accepted HTTP connection from: " +
                     "\(clientSocket.remoteHostname):\(clientSocket.remotePort)")
 				
-                DispatchQueue.global().async { [weak self] in
-                    if let strongSelf = self {
-                        strongSelf.initializeClientConnection(clientSocket: clientSocket, listenSocket: listenSocket, socketManager: socketManager)
+                if let _ = listenSocket.delegate {
+                    DispatchQueue.global().async { [weak self] in
+                        if let strongSelf = self {
+                            strongSelf.initializeClientConnection(clientSocket: clientSocket, listenSocket: listenSocket)
+                            strongSelf.handleClientConnection(clientSocket: clientSocket, socketManager: socketManager)
+                        }
                     }
+                } else {
+                    handleClientConnection(clientSocket: clientSocket, socketManager: socketManager)
                 }
-                
             } catch let error {
                 if self.state == .stopped {
                     if let socketError = error as? Socket.Error {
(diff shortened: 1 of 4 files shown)


youming-lin:
@djones6 
Consistent errors from Travis tests. Can you investigate?

djones6:
@youming-lin I can recreate the failures locally - I'm puzzled as to why I didn't see these before...  Investigating.

djones6:
Okay, I think the reason I didn't see them before was that some additional checks have been added on top of my original PR for Socket:  https://github.com/IBM-Swift/BlueSocket/commit/5c77668e3159ec345b23ca9de4aecc46919d56f2

This introduces a check to see whether the call to `invokeDelegateOnAccept` is appropriate.  The errors being thrown during the tests are `-9966` which is the new `SOCKET_ERR_INVALID_DELEGATE_CALL`.  So the problem is apparently related to how we are setting the socket's delegate (perhaps these extra checks have highlighted a bug).  Still investigating.

billabt:
That exception is thrown under two conditions. The first is when the API is called with a Socket that was not returned by an _acceptClientConnection(invokeDelegate: false)_. The second instance is when it's called multiple times for the same Socket.

djones6:
The exception in this case is because the socket does not have a delegate.  The original change I made to Socket allowed `invokeDelegateOnAccept` to be invoked regardless of whether a delegate was present (if `delegate` was `nil`, it simply did nothing).

The check does not allow this, as `self.delegate` is nil, and so `needsAcceptDelegateCall` is never set to `true` (regardless of how `acceptClientConnection` was called).

Kitura-net now needs to check whether the socket has a delegate before trying to invoke the new function.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -192,12 +192,16 @@ public class HTTPServer: Server {
                 Log.debug("Accepted HTTP connection from: " +
                     "\(clientSocket.remoteHostname):\(clientSocket.remotePort)")
 				
-                DispatchQueue.global().async { [weak self] in
-                    if let strongSelf = self {
-                        strongSelf.initializeClientConnection(clientSocket: clientSocket, listenSocket: listenSocket, socketManager: socketManager)
+                if let _ = listenSocket.delegate {
```
ianpartridge: `if listenSocket.delegate { }` 🙂

djones6:
Through fixing the bug above, I realized that I had added an async invocation for every socket accept, even if there is no delegate (which is unnecessary).

I pushed a refactoring that only goes down the async route if the socket has a delegate.  I think it also makes it clearer which parts of the process the error handling applies to.

djones6:
I'm adding a test to cover the original bug.

PR #221, opened 2017-10-13
Avoid calling handleClientConnection if initializeClientConnection fails

djones6:
## Description

This is a follow-on to #219, which split out the socket accept from the protocol initialization for SSL connections, in order to perform the latter on a separate thread.

In the case where the socket initialization fails, such as with an SSL handshake error, following #219 we would log the error but then attempt to handle the client connection anyway, resulting in unnecessary work and further errors.

This PR moves the `do`/`catch` logic for `initializeClientConnection` so that we only try to call `handleClientConnection` if the former was successful.

I also updated the log messages to document the source of a connection when it fails initialization.

## Motivation and Context

Avoids unnecessary attempts to handle a connection for which the SSL initialization has failed, and reporting of further errors as a consequence.

## How Has This Been Tested?

I have run the Kitura-net tests and they are passing.  Note that there is an intermittent test failure exposed by the test intrduced in #219 - which intentionally generates a bad connection to an SSL listener - which will be resolved by https://github.com/IBM-Swift/BlueSSLService/pull/42.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServer.swift
+++ b/Sources/KituraNet/HTTP/HTTPServer.swift
@@ -189,14 +189,29 @@ public class HTTPServer: Server {
         repeat {
             do {
                 let clientSocket = try listenSocket.acceptClientConnection(invokeDelegate: false)
-                Log.debug("Accepted HTTP connection from: " +
-                    "\(clientSocket.remoteHostname):\(clientSocket.remotePort)")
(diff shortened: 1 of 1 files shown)


djones6:
I'm accepting the apparent reduction in coverage - this is really because of a fix I added to BlueSSLService which no longer (erroneously) triggers an error handling path.  Raised #222 for a proper test for this path.

PR #225, opened 2017-10-23
Fix parsing of responses to HEAD requests with non-zero Content-Length

tunniclm:
## Description
I have added a new parameter `skipBody: Bool` to both `ClientReponse.init()` and `HTTPParser.init()` which tells the response to skip parsing of the request body.  These parameters default to `false`.

When the `ClientRequest` creates the `ClientReponse` object to represent the response from the server, it now checks if the `method` is `HEAD` and if so passes `skipBody: true` to the `ClientResponse` constructor. This will be passed down to the `HTTPParser` constructor (this required a small change to the way the `httpParser` property is initialized) where a decision is made on how to initialize the callbacks to the CHTTPParser. If `skipBody` is `false` then the callback is created exactly as before, however, if `skipBody` is `true` then the callback is created so that it returns `1` which tells `CHTTPParser` not to parse the body content.

I was not able to find a better way to communicate this flag to `CHTTPParser` since it clears all the parser `flags` when the parse step begins.

## Motivation and Context
KituraNet client responses to HEAD requests do not properly handle having a non-zero "Content-Length" header. According to the HTTP specification HEAD requests should have identical headers to the corresponding GET request on a given URL -- this includes the "Content-Length" header. It also says that servers MUST NOT provide any body content for responses to HEAD requests.

KituraNet client doesn't handle this scenario properly and the HEAD test server it runs against incorrectly sets the "Content-Length" to 0 when responding to a HEAD request which makes the test pass.

## How Has This Been Tested?
I modified the KituraNet client HEAD tests so the test server sets the "Content-Length" appropriately (which causes the test to fail without the other changes in this PR).

The changes in this PR pass the KituraNet and Kitura tests.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTPParser/HTTPParser.swift
+++ b/Sources/KituraNet/HTTPParser/HTTPParser.swift
@@ -63,9 +63,10 @@ class HTTPParser {
     /// Initializes a HTTPParser instance
     ///
     /// - Parameter isRequest: whether or not this HTTP message is a request
+    /// - Parameter skipBody: whether parser should skip body content (ie when parsing the response to a HEAD request)
     ///
     /// - Returns: an HTTPParser instance
-    init(isRequest: Bool) {
+    init(isRequest: Bool, skipBody: Bool = false) {
 
         self.isRequest = isRequest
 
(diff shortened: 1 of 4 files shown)


Sources/KituraNet/HTTPParser/HTTPParser.swift
```diff
@@ -99,14 +100,35 @@ class HTTPParser {
...
-            let results = getResults(parser)
-            
-            results?.onHeadersComplete(method: method, versionMajor: (parser?.pointee.http_major)!,
-                versionMinor: (parser?.pointee.http_minor)!)
-            return 0
+                let results = getResults(parser)
+                
+                results?.onHeadersComplete(method: method, versionMajor: (parser?.pointee.http_major)!,
+                    versionMinor: (parser?.pointee.http_minor)!)
+
```
shmuelk: The above lines of code should be extracted into a static/fileprivate function just getResults is.

The second group of similar lines of code below, should call the same function.

PR #227, opened 2017-10-26
Prevent collision on HTTPParser state by consecutive keepalive requests

### LINKED ISSUE #226: EXC_BAD_ACCESS when reading HTTPServerRequest.method with SwiftMetricsKitura
I've encountered an intermittent crash when running load against a Kitura server with SwiftMetricsKitura HTTP monitoring enabled.  There seem to be two distinct types of crash, though I believe their root cause may be related.

1) A crash happens frequently when SwiftMetrics is attempting to access the [`HTTPServerRequest.method`](https://github.com/IBM-Swift/Kitura-net/blob/1.7.18/Sources/KituraNet/HTTP/HTTPServerRequest.swift#L41) value in [`SwiftMetricsKitura.finished()`](https://github.com/RuntimeTools/SwiftMetrics/blob/1.2.4%40swift-4/Sources/SwiftMetricsKitura/SwiftMetricsKitura.swift#L73).

To make it clearer which field is causing the issue, I added a little extra debug (below) to access the field outside of the `HTTPData` constructor:
screen shot 2017-10-24 at 10 00 12

2) A crash in `HeadersContainer.append()` where we try to free a pointer that has already been freed.
screen shot 2017-10-24 at 10 23 36

Both crashes can be eliminated by avoiding reuse of the HTTPParser, instead allocating a new one in [`IncomingHTTPSocketProcessor.process()`](
https://github.com/IBM-Swift/Kitura-net/blob/1.7.18/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift#L89).

So I believe the corruption / bad behaviour is somehow related to the lifecycle of the HTTP parser.  I'm currently of the opinion that SwiftMetrics itself is behaving correctly, but is exposing an issue in Kitura-net by retaining a reference to the HTTPServerRequest beyond the point at which the response has been sent.

djones6:
## Description

~~Adds a lock around the use of the `HTTPParser`'s state~~ 
Move the `keepAlive()` call in `HTTPServerRequest` to be the last action taken during `end()`, to prevent data associated with an `HTTPServerRequest` (such as the `method`) from being overwritten by a subsequent request on the same Keep-Alive connection before `end()` completes.

## Motivation and Context

Resolves #226 

This fixes a timing window which is exposed by the SwiftMetricsKitura HTTP monitoring.

As soon as the response has been sent, the `IncomingHTTPSocketProcessor`'s state - and the `HTTPParser` associated with it - are reset, ready to handle a subsequent request from the same client.

The monitoring `finished()` events are processed synchronously on a serial queue, after a response has been sent to the client. However, because the entire handling of a request is performed on a separate thread to the socket reader, the reader is free to start processing a new request as soon as `keepAlive()` is called.

There is then a race between the client receiving the response and immediately sending the next request, and the monitoring processing the event that was queued (which will read values such as the `method` from the `HTTPServerRequest`).  Occasionally, the server will crash because the `ParseResults` is being mutated for the new request whilst it is still associated with the previous one (two `HTTPServerRequest` instances exist with a reference to the same `HTTPParser`).

Another change we could consider would be to copy the values needed for monitoring, then perform the monitoring asynchronously. This would add some small overhead, but would allow a subsequent request to be processed without waiting for monitoring to complete.

## How Has This Been Tested?

I have run the Kitura-net tests on Linux and Mac.

I've also verified that this resolves the intermittent crash with SwiftMetricsKitura on Linux and Mac.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -120,7 +120,6 @@ public class HTTPServerResponse : ServerResponse {
             // Ordering is important here. Keepalive allows the processor to continue
             // processing further requests, so must only be called once monitoring
             // has completed, as the HTTPParser for this connection is reused.
-            processor.handlingComplete()
             if  keepAlive {
                 processor.keepAlive()
             }
--- a/Sources/KituraNet/HTTP/HTTPServerResponse.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerResponse.swift
@@ -107,22 +107,22 @@ public class HTTPServerResponse : ServerResponse {
     public func end() throws {
         if let processor = processor {
             try flushStart()
-            
-            let keepAlive = processor.isKeepAlive
-            
-            if  keepAlive {
-                processor.keepAlive()
-            }
-            
             if  buffer.length > 0  {
                 processor.write(from: buffer)
             }
-            
+            let keepAlive = processor.isKeepAlive
             if !keepAlive && !processor.isUpgrade {
                 processor.close()
             }
             if let request = request {
                 Monitor.delegate?.finished(request: request, response: self)
+            }
+            // Ordering is important here. Keepalive allows the processor to continue
+            // processing further requests, so must only be called once monitoring
+            // has completed, as the HTTPParser for this connection is reused.
+            processor.handlingComplete()
+            if  keepAlive {
+                processor.keepAlive()
             }
         }
     }
--- a/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
+++ b/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
@@ -123,6 +123,7 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
         keepAliveUntil=0.0
         inProgress = false
         clientRequestedKeepAlive = false
+        parserUnlock()  // In case the socket was closed while we were handling a request
         handler?.prepareToClose()
     }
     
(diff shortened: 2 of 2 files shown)


shmuelk:
I don't understand how this PR solves the fundamental problem.

The fundamental problem is that some outside entity, the monitoring code in this case, is saving a reference to the ServerRequest and ServerResponse after Kitura-net thinks it is finished with them and can do what it wants with them.

You said in your description that the monitoring code handles the request in another serial queue.
Thus locking inside of Kitura-net will not fix the problem.

Moving the call to processor.keepAlive() after the call to the monitor is a good thing. It would enable the monitor code to peak in the Kitura-net onbjects and **copy** the pieces of information it needs to perform its work. That copied information, actually copied values and not references, needs to be placed on the monitor's event queue. When from a Kitura-net perspective a request has finished, Kitura-net should be free to do what ever it wants with those objects.

Additionally using locks on the HTTP Parser is the wrong thing to do. First of all it doesn't solve the fundamental problem as you aren't waiting for the monitor to somehow tell you that it is done with the ServerRequest and ServerResponse objects. Second of all it will block threads all over the place needlessly. The HTTP parsing is done on the thread that does the socket I/O. On Linux we have a small set of threads that does this. Even on MacOS where this is done on the Global queue, there is a limit to the number of threads available.

While the discussion above mentions the race condition that exists between the client sending new requests and the objects being finished with, pipelining makes this even worse. As the data is already in the server waiting to be processed.

djones6:
@shmuelk It seems the locking was unnecessary.  Having gone back to the start, simply reordering the `keepAlive()` call so that it is the last act of `end()` is sufficient to fix the problem.

I've pushed a commit to back out the locking.

For clarification, whilst the monitoring is done on a separate serial queue, it is done synchronously.  `end()` does not complete until the monitoring completes.  However, what you said about copying the data would still offer an advantage: we could then queue the monitoring asynchronously, and continue processing further requests from the client immediately.  But we'd need to be careful with memory footprint if the monitoring queue were to become backlogged.

shmuelk:
Now it looks good.

djones6:
@tunniclm This is ready to merge, if you are happy with it.

PR #229, opened 2017-11-01
Support HTTP Pipelining

djones6:
## Description

Allows Kitura-net to handle pipelined requests, by pausing the `http_parser` when the `on_message_complete` callback is triggered.  This causes the `http_parser_execute` function to return, rather than continuing to process the rest of the buffer, and allows the processor to handle the first request.

Upon the completion of the response, `keepAlive()` is called, followed by `processBufferedReadData()`, which resets the parser state, and resumes parsing the buffer.

**Note** - On Linux, with the default (epoll) I/O implementation, pipelining is now functional but performs very slowly.  It works (and performs) correctly with GCD both on Linux and Mac.  I've raised a separate PR #230 to resolve it.  Performance is unaffected when pipelining is not used.

## Motivation and Context

Resolves IBM-Swift/Kitura#932

HTTP Pipelining support is required by the [TechEmpower Benchmarks](https://www.techempower.com/benchmarks/), which uses it during the `plaintext` test.  This is therefore a prerequisite to a correctly functioning Kitura TechEmpower implementation.

## How Has This Been Tested?

I ran the Kitura-net tests on Mac and Linux locally, and they passed.
I used `wrk` to test pipelining, as follows:
```
wrk -c128 -t16 -d10s -i1s http://localhost:8080/plaintext --script pipeline.lua 16 --
```
The `pipeline.lua` script for `wrk` is [generated by the TechEmpower benchmarks project](https://github.com/TechEmpower/FrameworkBenchmarks/blob/70a2dac6639cf731dc7c5730333d2ce4116c39b5/toolset/setup/linux/client.sh#L33) - a copy is available here: https://github.com/djones6/Swift-Bench/blob/master/pipeline.lua

I have added a couple of E2E tests for pipelining.  Since `ClientRequest` does not support the notion of pipelining, I have done this using Socket directly with hand-crafted requests.  We are still able to use the `ClientResponse` to inspect the results.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTPParser/HTTPParser.swift
+++ b/Sources/KituraNet/HTTPParser/HTTPParser.swift
@@ -117,8 +117,16 @@ class HTTPParser {
             }
         }
         
+        // When the parser reaches the end of a message, we will pause
+        // so that execute() completes and the handler can be invoked.
+        // Any remaining data in the buffer will be consumed as follows:
+        // - if Keep-Alive is disabled, the data will be discarded.
+        // - if Keep-Alive is enabled, the keepAlive() call and the
+        //   subsequent handleBuffererdReadData() call will reset the
+        //   parser's state, and then resume parsing the buffer.
         settings.on_message_complete = { (parser) -> Int32 in
             getResults(parser)?.onMessageComplete()
+            http_parser_pause(parser, 1)
             return 0
         }
         
--- a/Tests/KituraNetTests/ClientE2ETests.swift
+++ b/Tests/KituraNetTests/ClientE2ETests.swift
@@ -15,6 +15,7 @@
  **/
 
 import Foundation
+import Dispatch
 
 import XCTest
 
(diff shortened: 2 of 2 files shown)


djones6:
I've established why performance (with epoll) on Linux is poor.  The design of our epoll loop only allows a single request to be processed per client per `epoll_wait()`, as the handler (run asynchronously) will not permit further requests to be processed until `keepAlive()` is called, and by that time, we'll already have reentered `epoll_wait()`.

This call has a timeout (hard-coded as 50ms) which explains why I get very low numbers of requests per second as soon as pipelining is enabled.  More explanation in #932.

djones6:
@tunniclm This is ready to merge, if you are happy with it.

I have a solution for the slow performance on Linux, but it introduces a dependency on another C library, so I think it belongs in its own PR.  From my analysis of what's causing it, it's safe to merge this as-is as it won't affect anyone other than clients who pipeline (who were completely broken before).

PR #230, opened 2017-11-03
Improve performance of HTTP Pipelining on Linux

djones6:
## Description

Adds an `eventfd` to the epoll descriptor which can be used to wake the `epoll_wait()` when it is time to process any buffered data (such as an additional pipelined request).  The `eventfd_write()` function is used to trigger an event on this fd.

**Note:** this requires IBM-Swift/CEpoll#2

## Motivation and Context

#229 allowed HTTP Pipelining to function properly, but on Linux, due to the design of our `epoll_wait()` loop, each pipelined request is delayed by around 50ms (epoll_wait timeout) before it is processed.  Further detail in IBM-Swift/Kitura#932.

By waking the `epoll_wait()` after the `keepAlive()` call, we can immediately process the buffered data.

We must wake the epoll thread rather than calling the `handleBufferedReadDataHelper()` directly, because we otherwise run the risk of accessing the `IncomingHTTPSocketProcessor` concurrently in two different threads (the handler thread which just called `keepAlive()`, and the epoll thread, which may wake up if the client sends more data).

## How Has This Been Tested?

I ran the Kitura-net tests, and they passed.
I ran the performance test described in #229 and verified the performance issue on Linux is resolved.

## Checklist:

### DIFF
--- a/Sources/KituraNet/IncomingSocketHandler.swift
+++ b/Sources/KituraNet/IncomingSocketHandler.swift
@@ -211,8 +211,10 @@ public class IncomingSocketHandler {
             }
         #endif
         #if !GCD_ASYNCH && os(Linux)
-            // Wake the epoll_wait thread early to allow it to process the buffer
-            eventfd_write(epollWakeFd, 0)
+            if readBuffer.length > 0 {
+                // Wake the epoll_wait thread early to allow it to process the buffer
+                eventfd_write(epollWakeFd, 0)
+            }
         #endif
     }
     
--- a/Sources/KituraNet/IncomingSocketHandler.swift
+++ b/Sources/KituraNet/IncomingSocketHandler.swift
@@ -21,6 +21,11 @@ import Foundation
 import LoggerAPI
 import Socket
 
+#if !GCD_ASYNCH && os(Linux)
+    import CEpoll
+    import Glibc
+#endif
+
 /// This class handles incoming sockets to the HTTPServer. The data sent by the client
 /// is read and passed to the current `IncomingDataProcessor`.
 ///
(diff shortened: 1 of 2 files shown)


djones6:
Performance results from Ubuntu 16.04, comparing with vs. without #229 and #230:
- `epoll` = linux default
- `gcd` = compiled with `-Xswiftc -DGCD_ASYNCH`
- `sm` = with SwiftMetrics HTTP request monitoring
- `fix` = with #229 and #230, but without using pipelining
- `pipe` = with #229 and #230, using pipelining (16 requests pipelined)

Workload is a simple `GET` with a `Hello, World!` plaintext response.
```
         epoll |    78966.3 |    82712.5 |    96.7 |        27716 |      1.6 |      2.5 |     17.7 |     5
     epoll_fix |    77223.7 |    80072.9 |    97.0 |        27764 |      1.6 |      2.7 |     18.4 |     5
    epoll_pipe |    84620.9 |    86668.5 |    98.5 |        28212 |      8.8 |        0 |    222.9 |     5

           gcd |    50237.0 |    51109.4 |    81.6 |        31167 |      2.5 |      3.6 |     16.6 |     5
       gcd_fix |    50683.2 |    51360.7 |    81.5 |        30553 |      2.5 |      3.7 |     19.4 |     5
      gcd_pipe |    53039.4 |    54430.9 |    76.4 |        30762 |     14.6 |        0 |     52.1 |     5

      epoll_sm |    37170.4 |    37717.7 |    74.7 |        36811 |      3.4 |      4.1 |     10.5 |     5
  epoll_sm_fix |    33789.0 |    34301.8 |    72.2 |        35748 |      3.7 |      4.5 |     14.5 |     5
 epoll_sm_pipe |    32557.9 |    33127.7 |    74.9 |        36504 |     31.7 |        0 |     78.6 |     5

        gcd_sm |    35288.0 |    35853.9 |    78.8 |        37929 |      3.6 |      4.2 |     22.7 |     5
    gcd_sm_fix |    34812.7 |    35441.3 |    78.6 |        29821 |      3.6 |      4.4 |     19.5 |     5
   gcd_sm_pipe |    40298.0 |    41110.8 |    75.6 |        38612 |     25.6 |        0 |     65.9 |     5
```

Couple of issues to investigate:
- epoll seems to be slightly slower with the fix than without. Possibly because we are waking the epoll_wait unnecessarily when there is no buffered data.  We may be able to check the buffer to see if we should wake epoll (if that's safe to do from another thread)
- epoll with SwiftMetrics monitoring is substantially slower with the fix

djones6:
Closing as the epoll + eventfd combination seems to have some unexpected and catastrophic performance problems that I have been unable to figure out.

Issue #228, opened 2017-11-01
EXC_BAD_INSTRUCTION on IncomingSocketHandler.swift line 197 (only Mac OS X)


zirinisp:
On our development server, which is running Mac OS X 10.12.6, we are getting EXC_BAD_INSTRUCTION. So far, it was not a huge problem, as we were only using OS X for development. We are currently looking at moving our production server from bluemix to OS X, so it is essential that we sort it out.

The crash happens after the server has run for a couple of hours. As the server is used for development the traffic is minimal. I am not 100% sure, but when the server crashes I believe that there is no traffic at all. So far, I have never managed to get it running for more than 1 day without getting the crash. The only way to replicate the crash is if I wait for a couple of hours. So far I have not managed to replicate it in any other way.

The crash happens on IncomingSocketHandler.swift line 197. The method that is causing the crash is handleBufferedReadData(). The method is only run when in OSX, iOS, tvOS and under GCD_ASYNC (which I am not sure what it means). That is probably the reason that we do not get it under Linux.

A screenshot from xcode has been attached for more information.

<img width="1348" alt="screen shot 2017-11-01 at 09 50 01" src="https://user-images.githubusercontent.com/2939569/32270194-c7f0f648-beec-11e7-870f-85c604da23d5.png">
<img width="1348" alt="screen shot 2017-10-31 at 17 33 11" src="https://user-images.githubusercontent.com/2939569/32270195-c81901d8-beec-11e7-8484-cb8c90902afb.png">

djones6:
I've been able to reproduce this locally, and am investigating a fix.

djones6:
@zirinisp I've tagged a fix for this issue in `Kitura-net 1.7.20`.  At least, it's a fix for the same symptom, and I'm fairly confident that this will turn out to be the issue you were having.  Details of what causes the problem are described in #232.

If you continue to see this crash after updating your level of Kitura-net, please let me know.

zirinisp:
Thank you very much. I will try it out and keep you posted if the issue comes back again.

cloutiertyler:
I'm able to semi-consistently reproduce what appears to be similar bug on `Kitura 2.0.0`. I am running multiple Kitura HTTP servers which are communicating with each other via HTTP. I cannot for the life of me seem to track down where this is coming from. From time to time I will get a double free in the Kitura parsing logic, but it's unclear if it's related. It seems to be causing bad accesses all over the place, but not all the time.

[...]

PR #232, opened 2017-11-10
Close Keep-Alive connections following a parser error

djones6:
## Description

Resolves a bug in Keep-Alive handling, which results in an infinite loop (with epoll) or a crash (with GCD) when a malformed request is sent on a Keep-Alive connection, after a previous message has been successfully processed.

## Motivation and Context
May address #228 

Recreate is as simple as:
```
curl -X GET -H 'Content-Length: 0' -H 'Connection: Keep-Alive' -d"abcd" http://localhost:8080/
```
This problem does not occur if `Connection: Close` is sent. In this situation, after handling the initial request, whether successful or not, we close the connection.

The problem also does not occur if the first message from the client is invalid (regardless of the `Connection` header). In this case, we immediately respond with `400 Bad Request` and close the connection.  This only happens due to ordering of when the `clientRequestedKeepAlive` flag is updated: we only set this flag to `true` once a request has been successfully parsed.

### Symptom
In the case of epoll, we repeatedly attempt to parse the remaining buffer as the epoll_wait times out every 50ms, and we have a handler in the list of handlers that require deferred processing.

In the case of GCD, we attempt to invoke `self.handleBufferedReadDataHelper()` as part of a direct call to `end()` when a parser error has occurred.  In this situation, we attempt to synchronously enqueue a block to the queue we are already executing on, which results in a crash.

### Solution
The solution presented here is to introduce a new piece of state: `parserErrored`, which is checked when evaluating `isKeepAlive`, in addition to `clientRequestedKeepAlive`.

## How Has This Been Tested?

I ran the Kitura-net tests locally on Linux and Mac.  I tested with the `curl` example above to verify that the problem is fixed.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
+++ b/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
@@ -199,6 +199,9 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
             Log.error("Failed to parse a request. \(parsingStatus.error!)")
             let response = HTTPServerResponse(processor: self, request: nil)
             response.statusCode = .badRequest
+            // We must avoid any further attempts to process data from this client
+            // after a parser error has occurred. (see Kitura-net#228)
+            clientRequestedKeepAlive = false
             do {
                 try response.end()
             }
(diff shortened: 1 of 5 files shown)


ianpartridge:
It's fantastic that you're fixing this stuff.  Can we add a test too?

PR #240, opened 2018-02-07
Allow POST success status codes besides 201 (CREATED)

c1ira:
Allow POST success status codes besides 201 (CREATED)

## Description
This change adds HTTPStatusCode extensions for identifying whether a status code represents success. The remainder of the change is PRed to Kitura: https://github.com/IBM-Swift/Kitura/pull/1218

## Motivation and Context
When Kitura receives a POST, its success response is 201 (CREATED). If one explicitly specifies any other success response, such as 200 (OK), Kitura treats it like an error.

## How Has This Been Tested?
Ran `swift test`.
Have been using this modification at Capital One for several months.

## Checklist:
- [x ] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)

### DIFF
--- a/Sources/KituraNet/HTTP/HTTP.swift
+++ b/Sources/KituraNet/HTTP/HTTP.swift
@@ -118,3 +118,15 @@ public enum HTTPStatusCode: Int {
     case unsupportedMediaType = 415, useProxy = 305, misdirectedRequest = 421, unknown = -1
     
 }
+
+extension HTTPStatusCode: Comparable {
+
+    public static func <(lhs: HTTPStatusCode, rhs: HTTPStatusCode) -> Bool { return lhs.rawValue < rhs.rawValue }
+
+}
+
+extension HTTPStatusCode {
+
+    public static var successRange: Range<HTTPStatusCode> { return .OK ..< .multipleChoices }
+
+}


Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -118,3 +118,15 @@ public enum HTTPStatusCode: Int {
...
+
+extension HTTPStatusCode: Comparable {
+
+    public static func <(lhs: HTTPStatusCode, rhs: HTTPStatusCode) -> Bool { return lhs.rawValue < rhs.rawValue }
+
+}
+
+extension HTTPStatusCode {
+
+    public static var successRange: Range<HTTPStatusCode> { return .OK ..< .multipleChoices }
```
djones6: It seems reasonable to define a range on `HTTPStatusCode` in this way.

djones6:
@c1ira Thanks for the PR.  I'm happy to merge this now in order to unblock work on IBM-Swift/Kitura#1218.

Ideally we'd have some tests to confirm that `successRange` represents the expected range, but I'm happy for that to be in a subsequent PR.

Sources/KituraNet/HTTP/HTTP.swift
ianpartridge: Yes, we might consider adding the other ranges while we're at it (1xx informational, 3xx redirection, 4xx clienterror, 5xx servererror)

c1ira:
Thanks, and I'll put the test on my to-do list.

PR #241, opened 2018-02-22
Updates for Swift 4.1

ShihabMehboob:
Updates to Swift 4.1, in reference to this issue: https://github.ibm.com/arf/planning-swift-solutions/issues/571


.swift-version
```diff
@@ -1 +1 @@
-4.0.3
+4.0.3
```
djones6: Unnecessary change, please can you remove this?

Package.swift
```diff
@@ -20,9 +20,9 @@
 import PackageDescription
 
 var dependencies: [Package.Dependency] = [
-    .package(url: "https://github.com/IBM-Swift/LoggerAPI.git", .upToNextMinor(from: "1.7.0")),
+    .package(url: "https://github.com/IBM-Swift/LoggerAPI.git", .upToNextMajor(from: "1.7.0")),
```
djones6: This is not right, as we tagged the LoggerAPI 4.1 PR at 2.0.0 (a new major).

As we go forward, we want to drop the use of `.upToNextMinor` and I personally think that `.upToNextMajor` is best avoided, because a) it's the default when you just specify `from:`, and b) it looks too similar to `.upToNextMinor` when scanning a Package.swift.  (maybe that's more a statement about my eyesight, but hey)

In summary I'd suggest simply `from: "2.0.0"`

Package.swift
```diff
@@ -20,9 +20,9 @@
...
     .package(url: "https://github.com/IBM-Swift/BlueSocket.git", .upToNextMinor(from: "0.12.0")),
-    .package(url: "https://github.com/IBM-Swift/CCurl.git", .upToNextMinor(from: "0.4.0")),
+    .package(url: "https://github.com/IBM-Swift/CCurl.git", .upToNextMinor(from: "0.5.0")),
     .package(url: "https://github.com/IBM-Swift/BlueSSLService.git", .upToNextMinor(from: "0.12.0"))
```
djones6: These are correct for now - we should discuss whether we want to re-tag CCurl as 1.0.0, since there's no reason for it to still be a pre-release.  If we decide to move to 1.0.0, we should update here (again, with `from:`)

Sources/KituraNet/IncomingSocketProcessor.swift
```diff
@@ -29,7 +29,11 @@ public protocol IncomingSocketProcessor: class {
 
     /// A back reference to the `IncomingSocketHandler` processing the socket that
     /// this `IncomingDataProcessor` is processing.
-    weak var handler: IncomingSocketHandler? { get set }
+    #if swift(>=4.1)
+        var handler: IncomingSocketHandler? { get set }
+    #else
+        weak var handler: IncomingSocketHandler? { get set }
+    #endif
```
djones6: Is this necessary for the code to still compile on Swift 4.0?  If so then fine, otherwise I would like to simplify this down to a single line (just dropping the `weak`).

Sources/KituraNet/BufferList.swift
```diff
@@ -90,7 +90,11 @@ public class BufferList {
         
         let result = min(length, localData.length - byteIndex)
         let bytes = localData.bytes.assumingMemoryBound(to: UInt8.self) + byteIndex
-        UnsafeMutableRawPointer(buffer).copyBytes(from: bytes, count: result)
+        #if swift(>=4.1)
+            UnsafeMutableRawPointer(buffer).copyMemory(from: bytes, byteCount: result)
+        #else
+            UnsafeMutableRawPointer(buffer).copyBytes(from: bytes, count: result)
+        #endif
```
djones6: Is this necessary for the code to still compile on Swift 4.0? If so then fine, otherwise I would like to simplify this down to a single line (just `copyMemory`).

tunniclm:
@djones6: since @ShihabMehboob is off today, can you update this PR in his stead so we can get it merged today? Thanks

djones6:
Superseded by #242.

PR #245, opened 2018-03-07
Add curl timeout to ClientRequest

Zig1375:
## Description
Add curl timeout in ClientRequest

## Motivation and Context
Sometime need it


ianpartridge:
Thanks for the PR!  Would this be better as a computed get/set property called `timeout`?

djones6:
@Zig1375 I'm closing as inactive.  Please see Ian's comment above and reopen if you still wish to contribute.

PR #246, opened 2018-03-12
Define additional ranges on HTTPStatusCode

djones6:
## Description

#240 recently introduced a `successRange` on HTTPStatusCode.  

This PR replicates the approach taken by the [swift-server working group](https://github.com/swift-server/HTTP), defining a computed enum that indicates the class (ie. range) that the status code falls within, defining `.informational`, `.successful`, `.redirection`, `.clientError` and `.serverError` status classes.

## How Has This Been Tested?

I have added a test for each class.

## Checklist:

### DIFF
--- a/Sources/KituraNet/HTTP/HTTP.swift
+++ b/Sources/KituraNet/HTTP/HTTP.swift
@@ -127,6 +127,19 @@ extension HTTPStatusCode: Comparable {
 
 extension HTTPStatusCode {
 
+    /// The range of HTTP status codes that represent informational responses (1xx)
+    public static var informationalRange: Range<HTTPStatusCode> { return .`continue` ..< .OK }
+
+    /// The range of HTTP status codes that represent success (2xx)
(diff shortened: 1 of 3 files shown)


ianpartridge:
I spoke to @seabaylea about this.  He suggested we take a look at what the HTTP working group did in this area.  They defined a `Class` enum: https://github.com/swift-server/http/blob/develop/Sources/HTTP/HTTPResponse.swift#L305

This would let us write code like `if statusCode.class == .success { }` rather than `if HTTPStatusCode.successRange.contains(statusCode) { }` which is more succinct.

djones6:
@ianpartridge I've updated the PR, please could you review again.  Thanks

djones6:
@c1ira Although we merged #240, we had a bit more of a think about this, and decided we'd prefer this approach for representing status code ranges.

If we go with this, it'll require a slight rework to your IBM-Swift/Kitura#1218 PR.  Does that seem reasonable to you?

c1ira:
Okay :+1:

PR #250, opened 2018-04-16
Update to Swift 4.1 and dependencies

ianpartridge:


Package.swift
```diff
@@ -21,9 +21,9 @@ import PackageDescription
 
 var dependencies: [Package.Dependency] = [
     .package(url: "https://github.com/IBM-Swift/LoggerAPI.git", .upToNextMinor(from: "1.7.0")),
-    .package(url: "https://github.com/IBM-Swift/BlueSocket.git", .upToNextMinor(from: "0.12.0")),
+    .package(url: "https://github.com/IBM-Swift/BlueSocket.git", from: "1.0.0"),
     .package(url: "https://github.com/IBM-Swift/CCurl.git", from: "1.0.0"),
-    .package(url: "https://github.com/IBM-Swift/BlueSSLService.git", .upToNextMinor(from: "0.12.0"))
+    .package(url: "https://github.com/IBM-Swift/BlueSSLService.git", from: "1.0.0")
```
djones6: If we are now happy to take minor updates to these repos (which I believe we are), any reason we would not also relax the minor constraint on LoggerAPI?

Package.swift
```diff
@@ -37,7 +37,7 @@ var kituraNetDependencies: [Target.Dependency] = [
 #if os(Linux)
 dependencies.append(contentsOf: [
     .package(url: "https://github.com/IBM-Swift/CEpoll.git", .upToNextMinor(from: "0.1.0")),
-    .package(url: "https://github.com/IBM-Swift/BlueSignals.git", .upToNextMinor(from: "0.9.0"))
+    .package(url: "https://github.com/IBM-Swift/BlueSignals.git", .upToNextMinor(from: "1.0.0"))
```
djones6: I think this should drop the `.upToNextMinor` to be consistent with the two above.  (not tolerating breaking changes in a minor)

djones6: We could also tag a 1.0.0 of CEpoll to complete the set and drop the minor constraint there too.

PR #256, opened 2018-04-24
Updated HTTP doc

KyeMaloy97:
Updated the HTTP class docs

### DIFF
--- a/Sources/KituraNet/HTTP/HTTP.swift
+++ b/Sources/KituraNet/HTTP/HTTP.swift
@@ -14,24 +14,41 @@
  * limitations under the License.
  */
 
-/* 
- * This source file includes content derived from the Swift.org Server APIs open source project
- * (http://github.com/swift-server/http)
- *
- * Copyright (c) 2017 Swift Server API project authors
- * Licensed under Apache License v2.0 with Runtime Library Exception
- *
- * See http://swift.org/LICENSE.txt for license information
- */
-
 import Foundation
 
 // MARK: HTTP
 
-/// A set of helpers for HTTP: status codes mapping, server and client request creation.
+/**
+A set of helpers for HTTP: status codes mapping, server and client request creation.
+
+### Usage Example: ###
+````swift
+ //Create a HTTP server.
+ let server = HTTP.createServer()
+ 
+ //Create a new a `ClientRequest` instance using a URL.
+ let request = HTTP.request("http://localhost/8080") {response in
+ ...
+ }
+ 
+ //Get a `ClientRequest` instance from a URL.
+ let getHTTP = HTTP.get("http://localhost/8080") { response in
+ ...
+ }
+ 
+ HTTP.escape(url: testString)
+````
+*/
 public class HTTP {
     
-    /// Mapping of integer HTTP status codes to the String description.
+    /**
+     Mapping of integer HTTP status codes to the String description.
+    
+    ### Usage Example: ###
+    ````swift
+     var statusText = HTTP.statusCodes[HTTPStatusCode.OK.rawValue]
+    ````
+    */
     public static let statusCodes = [
         
         100: "Continue", 101: "Switching Protocols", 102: "Processing",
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -14,24 +14,41 @@
  * limitations under the License.
  */
 
-/* 
- * This source file includes content derived from the Swift.org Server APIs open source project
```
helenmasters: This needs copyright needs adding back in.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -14,24 +14,41 @@
...
 import Foundation
 
 // MARK: HTTP
 
-/// A set of helpers for HTTP: status codes mapping, server and client request creation.
+/**
+A set of helpers for HTTP: status codes mapping, server and client request creation.
+
+### Usage Example: ###
+````swift
```
helenmasters: The indenting in the example below needs fixing, i.e. the ... should be indented.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -14,24 +14,41 @@
...
+ //Create a HTTP server.
+ let server = HTTP.createServer()
+ 
+ //Create a new a `ClientRequest` instance using a URL.
+ let request = HTTP.request("http://localhost/8080") {response in
```
helenmasters: This needs a space adding before the word `response` for readability.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -14,24 +14,41 @@
...
+ ...
+ }
+ 
+ //Get a `ClientRequest` instance from a URL.
+ let getHTTP = HTTP.get("http://localhost/8080") { response in
+ ...
+ }
+ 
+ HTTP.escape(url: testString)
```
helenmasters: We should have a comment for what escape does (as per the other examples) and define an example `testString` as otherwise the user will just get compiler warnings.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
...
-    /// - Parameter url: URL address for the request.
-    /// - Parameter callback: closure to run after the request.
-    /// - Returns: a ClientRequest instance.
-    ///
-    /// - Note: This method will invoke the end function of the `ClientRequest`
-    ///        immediately after its creation.
+    /**
+    Get a `ClientRequest` using URL.
+    
+    - Parameter url: URL address for the request.
```
helenmasters: - Parameter url: The URL address for the request.
- Parameter callback: The closure to run after the request.
- Returns: A ClientRequest instance.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
...
     
-    /// Create a new `ClientRequest` using a list of options.
-    ///
-    /// - Parameter options: a list of `ClientRequest.Options`.
-    /// - Parameter callback: closure to run after the request.
-    /// - Returns: a `ClientRequest` instance
+    /**
+    Create a new `ClientRequest` using a list of options.
+    
+    - Parameter options: a list of `ClientRequest.Options`.
```
helenmasters: - Parameter options: A list of `ClientRequest.Options`.
- Parameter callback: The closure to run after the request.
- Returns: A `ClientRequest` instance.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
         511: "Network Authentication Required"
     ]
     
-    /// Create a new `HTTPServer`.
-    ///
-    /// - Returns: an instance of `HTTPServer`.
+    /**
+    Create a new `HTTPServer`.
+    
+    - Returns: an instance of `HTTPServer`.
```
helenmasters: an needs a capital letter -> An instance of `HTTPServer`.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
...
+    /**
+    Create a new `ClientRequest` using URL.
+    
+    - Parameter url: URL address for the request.
+    - Parameter callback: closure to run after the request.
+    - Returns: a `ClientRequest` instance
+    
+    ### Usage Example: ###
+    ````swift
+     let request = HTTP.request("http://localhost/8080") {response in
```
helenmasters: Need a space between  `{response`

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -146,22 +304,22 @@ extension HTTPStatusCode {
...
+            case 200..<300: self = .successful
+            case 300..<400: self = .redirection
+            case 400..<500: self = .clientError
+            case 500..<600: self = .serverError
+            default: self = .invalidStatus
             }
         }
     }
-
+
```
helenmasters: Please can you put a full stop at the end of the following
/// The `Class` representing the class of status code for this response status

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -110,27 +168,127 @@ public class HTTP {
...
+    /// HTTP code -1
+    unknown = -1
     
 }
 
-extension HTTPStatusCode {
 
+extension HTTPStatusCode {
+    
     /// The class of a `HTTPStatusCode` code
```
helenmasters: Please can you put a full stop at the end of the following
/// The class of a `HTTPStatusCode` code

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
...
-    /// Create a new `ClientRequest` using URL.
-    ///
-    /// - Parameter url: URL address for the request.
-    /// - Parameter callback: closure to run after the request.
-    /// - Returns: a ClientRequest instance.
-    ///
-    /// - Note: This method will invoke the end function of the `ClientRequest`
-    ///        immediately after its creation.
+    /**
+    Get a `ClientRequest` using URL.
```
helenmasters: I think we should add the following to the description here - "The only difference between this method and `HTTP.request()` is that it invokes the end function of the `ClientRequest` automatically." 

My reasoning is that if you look at the Node.js documentation it provides an http.get() function but in that case it sets the method to GET and calls req.end() automatically. Looking at our implementation I don't believe that we are setting the method type to GET here so I don't want users to get confused between the behaviour of Swift and Node.

Sources/KituraNet/HTTP/HTTP.swift
```diff
@@ -52,39 +69,73 @@ public class HTTP {
...
-    /// Create a new `ClientRequest` using URL.
-    ///
-    /// - Parameter url: URL address for the request.
-    /// - Parameter callback: closure to run after the request.
-    /// - Returns: a `ClientRequest` instance
+    /**
+    Create a new `ClientRequest` using URL.
+    
+    - Parameter url: URL address for the request.
+    - Parameter callback: closure to run after the request.
```
helenmasters: I think the wording (which is written a few times in this file) "The closure to run after the request" could be clearer... do we mean the closure to run when the `response` event is received?

PR #258, opened 2018-04-24
Updated HTTPServerRequest docs

KyeMaloy97:
Updated HTTPServerRequest docs with new code examples and improvements

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -20,11 +20,37 @@ import LoggerAPI
 
 // MARK: HTTPServerRequest
 
-/// This class implements the `ServerRequest` protocol for incoming sockets that
-/// are communicating via the HTTP protocol. 
+/**
+This class implements the `ServerRequest` protocol for incoming sockets that are communicating via the HTTP protocol. Data and Strings can be read in.
+
+### Usage Example: ###
+````swift
+ func handlePost(request: ServerRequest, response: ServerResponse) {
+     var body = Data()
+     do {
+         let length = try request.readAllData(into: &body)
+         let result = "Read \(length) bytes"
+         response.headers["Content-Type"] = ["text/plain"]
+         response.headers["Content-Length"] = ["\(result.count)"]
+ 
+         try response.end(text: result)
+     }
+     catch {
+         print("Error reading body or writing response")
+     }
+ }
+````
+*/
 public class HTTPServerRequest: ServerRequest {
     
-    /// HTTP Status code if this message is a response
+    /**
+     HTTP Status code if this message is a response
+     
+     ### Usage Example: ###
+     ````swift
+     response!.httpStatusCode
+     ````
+     */
     @available(*, deprecated, message:
     "This method never worked on Server Requests and was inherited incorrectly from a super class")
     public private(set) var httpStatusCode: HTTPStatusCode = .unknown
@@ -32,28 +58,78 @@ public class HTTPServerRequest: ServerRequest {
     /// Client connection socket
     private let socket: Socket
     
-    /// Server IP address pulled from socket.
+    /**
+     Server IP address pulled from socket.
+     
+     ### Usage Example: ###
+     ````swift
+     request.remoteAddress
+     ````
+     */
     public var remoteAddress: String {
         return socket.remoteHostname
     }
     
-    /// HTTP Method of the request.
+    /**
+     HTTP Method of the request.
+     
+     ### Usage Example: ###
+     ````swift
+     request.method.lowercased()
+     ````
+     */
     public var method: String { return httpParser?.method ?? ""}
     
-    /// Major version of HTTP of the request
+    /**
+     Major version of HTTP of the request
+     
+     ### Usage Example: ###
+     ````swift
+     print(String(describing: request.httpVersionMajor))
+     ````
+     */
     public var httpVersionMajor: UInt16? { return httpParser?.httpVersionMajor ?? 0}
     
-    /// Minor version of HTTP of the request
+    /**
+     Minor version of HTTP of the request
+     
+     ### Usage Example: ###
+     ````swift
+     print(String(describing: request.httpVersionMinor))
+     ````
+     */
     public var httpVersionMinor: UInt16? { return httpParser?.httpVersionMinor ?? 0}
     
-    /// Set of HTTP headers of the request.
+    /**
+     Set of HTTP headers of the request.
+     
+     ### Usage Example: ###
+     ````swift
+     let protocols = request.headers["Upgrade"]
+     ````
+     */
     public var headers: HeadersContainer { return httpParser?.headers ?? HeadersContainer() }
     
-    /// socket signature of the request.
+    /**
+     socket signature of the request.
+     
+     ### Usage Example: ###
+     ````swift
+     public var signature: Socket.Signature? { return socket.signature }
+     ````
+     */
     public var signature: Socket.Signature? { return socket.signature }
     
     private var _url: URL?
     
+    /**
+     Create and validate the full URL.
+     
+     ### Usage Example: ###
+     ````swift
+     print(request.urlURL)
+     ````
+     */
     public var urlURL: URL {
         if let _url = _url {
             return _url
@@ -90,8 +166,15 @@ public class HTTPServerRequest: ServerRequest {
     
     private var urlc: URLComponents?
     
-    /// The URL from the request as URLComponents
-    /// URLComponents has a memory leak on linux as of swift 3.0.1. Use 'urlURL' instead
+    /**
+     The URL from the request as URLComponents
+     URLComponents has a memory leak on linux as of swift 3.0.1. Use 'urlURL' instead
+     
+     ### Usage Example: ###
+     ````swift
+     print(request.urlComponents)
+     ````
+     */
     @available(*, deprecated, message:
     "URLComponents has a memory leak on linux as of swift 3.0.1. use 'urlURL' instead")
     public var urlComponents: URLComponents {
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
 
 // MARK: HTTPServerRequest
 
-/// This class implements the `ServerRequest` protocol for incoming sockets that
-/// are communicating via the HTTP protocol. 
+/**
+This class implements the `ServerRequest` protocol for incoming sockets that are communicating via the HTTP protocol. Data and Strings can be read in.
+
+### Usage Example: ###
```
helenmasters: We should have something here to describe this example even if it's just something like
"In this example, we read the request data from the body of `request` and return the length of the data in bytes in the `response` object."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+ }
+````
+*/
 public class HTTPServerRequest: ServerRequest {
     
-    /// HTTP Status code if this message is a response
+    /**
+     HTTP Status code if this message is a response
+     
+     ### Usage Example: ###
```
helenmasters: We should remove this example as this method is marked as deprecated, so we don't want to encourage users to use it.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ````swift
+     response!.httpStatusCode
+     ````
+     */
     @available(*, deprecated, message:
```
helenmasters: jazzy doesn't support generating documentation for @available (https://github.com/realm/jazzy/issues/843)

Therefore, as the current message is not very readable in the API docs can we make this a jazzy note, or does this need to be marked @deprecated to display as such within Xcode? Regardless please can we change the text to:
"This method does not work on ServerRequest objects. It was incorrectly inherited from a super class."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+         print("Error reading body or writing response")
+     }
+ }
+````
+*/
 public class HTTPServerRequest: ServerRequest {
     
-    /// HTTP Status code if this message is a response
+    /**
+     HTTP Status code if this message is a response
```
helenmasters: How can this be a response object when we're inside ServerRequest?

helenmasters: Please add a comma and a full stop -> "HTTP Status code, if this message is a response."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
     @available(*, deprecated, message:
     "This method never worked on Server Requests and was inherited incorrectly from a super class")
     public private(set) var httpStatusCode: HTTPStatusCode = .unknown
     
     /// Client connection socket
     private let socket: Socket
     
-    /// Server IP address pulled from socket.
+    /**
+     Server IP address pulled from socket.
```
helenmasters: Suggested reword -> "The server IP address from the socket."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     
+     ### Usage Example: ###
+     ````swift
+     request.remoteAddress
```
helenmasters: We should at least assign this to a variable or constant. We could also print out an example, something like that below (but the second piece is optional)

```
let remoteAddress = request.remoteAddress
print(remoteAddress) // Prints xx.xx.xx.xx
```

(where xx.xx.xx.xx is an example real IP address)

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ````
+     */
     public var remoteAddress: String {
         return socket.remoteHostname
     }
     
-    /// HTTP Method of the request.
+    /**
+     HTTP Method of the request.
```
helenmasters: The HTTP method of the request.
(Method needs to be lowercase).

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     
+     ### Usage Example: ###
+     ````swift
+     request.method.lowercased()
```
helenmasters: Again assign to a constant or variable and ideally print out an example.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ````
+     */
     public var method: String { return httpParser?.method ?? ""}
     
-    /// Major version of HTTP of the request
+    /**
+     Major version of HTTP of the request
```
helenmasters: "The HTTP major version of the request."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ### Usage Example: ###
+     ````swift
+     print(String(describing: request.httpVersionMajor))
+     ````
+     */
     public var httpVersionMajor: UInt16? { return httpParser?.httpVersionMajor ?? 0}
     
-    /// Minor version of HTTP of the request
+    /**
+     Minor version of HTTP of the request
```
helenmasters: "The HTTP minor version of the request."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ### Usage Example: ###
+     ````swift
+     print(String(describing: request.httpVersionMinor))
+     ````
+     */
     public var httpVersionMinor: UInt16? { return httpParser?.httpVersionMinor ?? 0}
     
-    /// Set of HTTP headers of the request.
+    /**
+     Set of HTTP headers of the request.
```
helenmasters: The word set is confusing in this description. Are we reading the headers in the request, or can we set them here too?

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     
+     ### Usage Example: ###
+     ````swift
+     let protocols = request.headers["Upgrade"]
```
helenmasters: Can we have a more obvious example than upgrade?

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     */
     public var headers: HeadersContainer { return httpParser?.headers ?? HeadersContainer() }
     
-    /// socket signature of the request.
+    /**
+     socket signature of the request.
+     
+     ### Usage Example: ###
+     ````swift
+     public var signature: Socket.Signature? { return socket.signature }
```
helenmasters: This is a strange choice for the usage example.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -20,40 +20,116 @@ import LoggerAPI
...
+     ### Usage Example: ###
+     ````swift
+     let protocols = request.headers["Upgrade"]
+     ````
+     */
     public var headers: HeadersContainer { return httpParser?.headers ?? HeadersContainer() }
     
-    /// socket signature of the request.
+    /**
+     socket signature of the request.
```
helenmasters: "The socket signature of the request."

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -90,8 +166,15 @@ public class HTTPServerRequest: ServerRequest {
     
     private var urlc: URLComponents?
     
-    /// The URL from the request as URLComponents
-    /// URLComponents has a memory leak on linux as of swift 3.0.1. Use 'urlURL' instead
+    /**
+     The URL from the request as URLComponents
+     URLComponents has a memory leak on linux as of swift 3.0.1. Use 'urlURL' instead
```
helenmasters: The first sentence needs a full stop at the end (as does the last)
swift -> Swift

We should remove the usage example if this is now deprecated.

As in the previous example above please try using a note rather than @available as this is not supported by jazzy

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -103,16 +186,30 @@ public class HTTPServerRequest: ServerRequest {
         return urlc
     }
     
-    /// The URL from the request in string form
-    /// This contains just the path and query parameters starting with '/'
-    /// Use 'urlURL' for the full URL
+    /**
+     The URL from the request in string form
```
helenmasters: Needs a full stop at the end and a comma after the word request.
You need to show an example full URL and then a urlString with just the path and query parameters to make this clearer.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -103,16 +186,30 @@ public class HTTPServerRequest: ServerRequest {
...
+     This contains just the path and query parameters starting with '/'
+     Use 'urlURL' for the full URL
+     
+     ### Usage Example: ###
+     ````swift
+     print(request.urlString)
+     ````
+     */
     @available(*, deprecated, message:
```
helenmasters: Why is this deprecated?

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -103,16 +186,30 @@ public class HTTPServerRequest: ServerRequest {
...
     "This contains just the path and query parameters starting with '/'. use 'urlURL' instead")
     public var urlString : String { return httpParser?.urlString ?? ""}
     
-    /// The URL from the request in UTF-8 form
-    /// This contains just the path and query parameters starting with '/'
-    /// Use 'urlURL' for the full URL
+    /**
+     The URL from the request in UTF-8 form
```
helenmasters: Needs full stops a the end of the sentences and a clear example.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -144,11 +241,18 @@ public class HTTPServerRequest: ServerRequest {
...
-    /// - Returns: the number of bytes read.
+    /**
+     Read a chunk of the body of the request.
+     
+     - Parameter into: An NSMutableData to hold the data in the request.
+     - Throws: if an error occurs while reading the body.
+     - Returns: the number of bytes read.
+     
+     ### Usage Example: ###
+     ````swift
```
helenmasters: All three of the read() examples are very short with no explanation.

The parameters/throws/returns definitions all need to start with capital letters.

helenmasters:
Approved, despite comments not being addressed. I may address the comments at a later date when I get chance, as they would be nice to have.

PR #265, opened 2018-04-24
Updated SPIUtils docs

KyeMaloy97:
Updated SPIUtils docs with new code examples


helenmasters:
Merging on the basis that it's better than the documentation that we currently have.

PR #266, opened 2018-04-24
Updated URLParser docs

KyeMaloy97:
Updated URLParser docs with new code examples

### DIFF
--- a/Sources/KituraNet/HTTPParser/URLParser.swift
+++ b/Sources/KituraNet/HTTPParser/URLParser.swift
@@ -27,29 +27,29 @@ import Foundation
 // MARK: URLParser
 
 /**
-A parsing of HTTP URL according to the folowing format:
+ Splits and parses URLs into components - scheme, host, port, path, query string etc. according to the following format:
 
-scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
+**scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]**
 
 ### Usage Example: ###
 ````swift
- //Initialize a new URLParser instance, and check whether or not a connection has been established.
- let url = "https://example.org/absolute/URI/with/absolute/path/to/resource.txt".data(using: .utf8)!
+ // Initialize a new URLParser instance, and check whether or not a connection has been established.
+ let url = "http://user:password@sample.host.com:8080/a/b/c?query=somestring#hash".data(using: .utf8)!
  let urlParser = URLParser(url: url, isConnect: false)
 ````
 */
 public class URLParser : CustomStringConvertible {
 
-    /// Schema.
+    /// The schema of the URL.
     public var schema: String?
 
-    /// Hostname.
+    /// The host component of the URL.
     public var host: String?
     
     /// Path portion of the URL.
     public var path: String?
     
-    /// The entire query portion of the URL.
+    /// The query component of the URL.
     public var query: String?
     
     /// An optional fragment identifier providing direction to a secondary resource.
@@ -62,11 +62,11 @@ public class URLParser : CustomStringConvertible {
     public var port: UInt16?
     
     /**
-    The query parameters broken out.
+    The value of the query component of the URL name/value pair, for the passed in query name.
     
     ### Usage Example: ###
     ````swift
-    let parsedURLParameters = urlParser.queryParameters["key"]
+    let parsedURLParameters = urlParser.queryParameters["query"]
     ````
     */
     public var queryParameters: [String:String] = [:]
@@ -109,10 +109,10 @@ public class URLParser : CustomStringConvertible {
     }
     
     /**
-    Initialize a new URLParser instance.
+    Initialize a new `URLParser` instance.
     
-    - Parameter url: URL to be parsed.
-    - Parameter isConnect: whether or not a connection has been established.
+    - Parameter url: The URL to be parsed.
+    - Parameter isConnect: A boolean, indicating whether or not a connection has been established.
     
     ### Usage Example: ###
     ````swift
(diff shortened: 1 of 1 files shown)


Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -85,11 +108,17 @@ public class URLParser : CustomStringConvertible {
         return desc
     }
     
+    /**
+    Initialize a new URLParser instance.
```
helenmasters: We should have backticks around `URLParser`

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -85,11 +108,17 @@ public class URLParser : CustomStringConvertible {
...
     
-    /// Initialize a new URLParser instance.
-    ///
-    /// - Parameter url: URL to be parsed.
-    /// - Parameter isConnect: whether or not a connection has been established.
+    - Parameter url: URL to be parsed.
+    - Parameter isConnect: whether or not a connection has been established.
```
helenmasters: A boolean, indicating whether or not a connection has been established.

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -26,9 +26,18 @@ import Foundation
 
 // MARK: URLParser
 
-/// A parsing of HTTP URL according to the folowing format:
-///   
-/// scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
+/**
+A parsing of HTTP URL according to the folowing format:
```
helenmasters: I'd prefer the wording "Splits and parses URLs into components - scheme, host, port, path, query string etc. according to the following format:"

I also suggest that you make the format string in bold so that it stands out more clearly.

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -26,9 +26,18 @@ import Foundation
...
+
+scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
+
+### Usage Example: ###
+````swift
+ //Initialize a new URLParser instance, and check whether or not a connection has been established.
+ let url = "https://example.org/absolute/URI/with/absolute/path/to/resource.txt".data(using: .utf8)!
```
helenmasters: Why don't we change the example to use the following string as then it will contain an entry for all of the components which we can pull out?
"http://user:password@sample.host.com:8080/a/b/c?query=somestring#hash"

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -26,9 +26,18 @@ import Foundation
...
-///   
-/// scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
+/**
+A parsing of HTTP URL according to the folowing format:
+
+scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
+
+### Usage Example: ###
+````swift
+ //Initialize a new URLParser instance, and check whether or not a connection has been established.
```
helenmasters: // Initialize

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -52,10 +61,24 @@ public class URLParser : CustomStringConvertible {
     /// The port specified, if any, in the URL.
     public var port: UInt16?
     
-    /// The query parameters broken out.
+    /**
+    The query parameters broken out.
+    
+    ### Usage Example: ###
+    ````swift
+    let parsedURLParameters = urlParser.queryParameters["key"]
```
helenmasters: If we change the example above it would make sense to reflect this here as 
`let parsedURLParameters = urlParser.queryParameters["query"]`

helenmasters: Please can we change /// The entire query portion of the URL. to
/// The query component of the URL.
I think we should have an example which shows that it would return "query=somestring" from the example above.

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -52,10 +61,24 @@ public class URLParser : CustomStringConvertible {
     /// The port specified, if any, in the URL.
     public var port: UInt16?
     
-    /// The query parameters broken out.
+    /**
+    The query parameters broken out.
```
helenmasters: /// The value of the query component of the URL name/value pair, for the passed in query name.

Sources/KituraNet/HTTPParser/URLParser.swift
KyeMaloy97: ✅

Sources/KituraNet/HTTPParser/URLParser.swift
KyeMaloy97: ✅

Sources/KituraNet/HTTPParser/URLParser.swift
```diff
@@ -26,9 +26,18 @@ import Foundation
...
+### Usage Example: ###
+````swift
+ //Initialize a new URLParser instance, and check whether or not a connection has been established.
+ let url = "https://example.org/absolute/URI/with/absolute/path/to/resource.txt".data(using: .utf8)!
+ let urlParser = URLParser(url: url, isConnect: false)
+````
+*/
 public class URLParser : CustomStringConvertible {
 
     /// Schema.
```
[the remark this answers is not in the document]
KyeMaloy97: ✅

Sources/KituraNet/HTTPParser/URLParser.swift
KyeMaloy97: ✅

Sources/KituraNet/HTTPParser/URLParser.swift
KyeMaloy97: ✅

KyeMaloy97:
Made the changes requested.

PR #267, opened 2018-05-01
Query parameters become nil on empty string

Andrew-Lees11:
## Description
To match the behavior is Kitura in this [pull request](https://github.com/IBM-Swift/Kitura/pull/1262), This pull request changes the Query parameter parser to treat empty strings as nil. 

## Motivation and Context
This was brought in following from an [issue](https://github.com/IBM-Swift/Kitura/issues/1261) In Kitura where HTML forms would send the empty string to represent nil for forms and we would claim they are invalid because they are strings. 

Since get forms use URL query parameters and to make our handling if URLEncoded data consistent we have decided to change our parsing of URLEncoded data to treat the empty string as nil. 

This parsing is done is done in both Kitura and Kitura-net and so there are two pull requests in implement these changes.

## How Has This Been Tested?
An extra test has been added which checks this behavior and all other tests are still passing on mac.

### DIFF
--- a/Sources/KituraNet/HTTPParser/URLParser.swift
+++ b/Sources/KituraNet/HTTPParser/URLParser.swift
@@ -149,7 +149,7 @@ public class URLParser : CustomStringConvertible {
             for pair in pairs {
                     
                 let pairArray = pair.components(separatedBy: "=")
-                if pairArray.count == 2 {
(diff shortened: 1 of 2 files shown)


Andrew-Lees11:
Closing in favour of new [PR](https://github.com/IBM-Swift/KituraContracts/pull/21/files) changing behavior inside Kitura Contracts.

PR #270, opened 2018-06-20
Issues related to ClientRequest and 3xx redirection

NocturnalSolutions:
I’ve discovered various glitches related to client redirection and ClientRequest.

- When doing a redirect, the same HTTP method will always be used for the new request; e.g., if a POST request returns a “Location” header, a POST request will be done to the new location. This is correct behavior when the response HTTP status code is 307 Temporary Redirect or 308 Permanent Redirect, and is correct inasmuch as the specifics of this are undefined for 301 Moved Permanently and 302 Found, but not correct for ~~308 Permanent Redirect~~ 303 See Other, for which the second request must explicitly be a GET request.
- Due to a `<` being used where a `<=` is needed, there is an off-by-one error in the code that counts whether the client has redirected more than `maxRedirects` times, such that the effective limit is actually `maxRedirects` minus one - setting `maxRedirects` to 1 will cause *no* redirects to be followed, just as with setting it to 0.
- If a “Location” header is present, `ClientRequest.prepareForRedirect()` is called, even if the `maxRedirects` limit has been hit and we’re not really going to do a redirect. As `prepareForRedirect()` wipes out the response data buffer, there’s no way to “read” the response to examine, for example, its response body or its headers. A particular use case where this might be desirable is to see if a route handler is returning a “Location” header in the first place.

## Description

This PR adds code to detect and properly handle the ~~308~~ 303 case, and to correct the above-mentioned bugs.

## Motivation and Context

I discovered these issues while trying to write a test for Kitura route handler that takes a user's POST data, creates a page with it, then redirects the user to the page's URL. My test failed because the content of the page after redirecting was an error page rather than what was expected, and it was because the client was doing a POST request to the path in the Location header when it should have been doing a GET one. Later experimentation and debugging surfaced the other issues.

## How Has This Been Tested?

The PR includes a new test case which checks various redirection-related behavior.

## Checklist:


ianpartridge:
Thanks for this!

> When doing a redirect, the same HTTP method will always be used for the new request; e.g., if a POST request returns a “Location” header, a POST request will be done to the new location. This is correct behavior when the response HTTP status code is 307 Temporary Redirect or 308 Permanent Redirect, and is correct inasmuch as the specifics of this are undefined for 301 Moved Permanently and 302 Found, but not correct for 308 Permanent Redirect, for which the second request must explicitly be a GET request.

Can you point me at a spec which says that the request following 308 Permanent Redirect must be a GET? I looked at https://tools.ietf.org/html/rfc7238 but this says "it does not allow changing the request method from POST to GET".

NocturnalSolutions:
(Sorry for the slow reply; not sure how I missed this.)

Sorry, I got my lines crossed in the OP. The status code I meant to say to force a GET request was [303 See Other](https://en.wikipedia.org/wiki/HTTP_303).

ianpartridge:
@djones6 could you check over this too please?

Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -381,6 +383,47 @@ class ClientE2ETests: KituraNetTest {
         }
     }
 
+    func testRedirection() {
+        performServerTest(TestRedirectDeletate()) { expectation in
```
djones6: Typo: `TestRedirectDeletate` -> `TestRedirectDelegate`

Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -427,6 +470,32 @@ class ClientE2ETests: KituraNetTest {
         }
     }
 
+    class TestRedirectDeletate: ServerDelegate {
```
djones6: Typo: `TestRedirectDeletate` -> `TestRedirectDelegate`

GarrettAlbright:
Oops. Pushed the typo fix.

NocturnalSolutions:
Oops. Just pushed the typo fix.

ianpartridge:
Thanks @NocturnalSolutions 👍

PR #273, opened 2018-07-20
Made the HTTPParser public

voynovia:
## Description
Made HTTPParser is public

## Motivation and Context
Extremely necessary

## How Has This Been Tested?
Build and Run tests )

### DIFF
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -236,7 +236,7 @@ public class HTTPServerRequest: ServerRequest {
     /// - Parameter httpParser: The `HTTPParser` object used to parse the incoming request
     ///
     /// - Returns: an HTTPServerRequest instance
-    init (socket: Socket, httpParser: HTTPParser?) {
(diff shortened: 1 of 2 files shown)


pushkarnk:
Just a thought. Given the requirement that we need to maintain the Kitura-Net and Kitura-NIO APIs identical, it may not be feasible to include an HTTPParser API in the latter.

voynovia:
Need for implementation of the Homekit Accessory Protocol.
https://github.com/Bouke/HAP

Maybe make the "HTTPParser" module to separate?

ianpartridge:
Hi @voynovia thanks for explaining. I'm afraid we don't want to expose more public API in Kitura-net as we are currently in the process of migrating Kitura from Kitura-net to Kitura-NIO. The HTTP parser in Kitura-net has a number of known limitations which mean we plan to deprecate it in future.

You are of course very welcome to fork this repository as you wish for your own purposes, but you might like to take a look at the swift-nio project which also provides an HTTP parser that may be suitable for your use.

Thanks for your PR and sorry we aren't able to merge it at this time.

PR #283, opened 2018-11-29
Build against Swift 5

djones6:
Adds testing with a Swift 5 development snapshot. The snapshot is defined in Travis as `SWIFT_DEVELOPMENT_SNAPSHOT` consistently across the IBM-Swift repos, to enable us to automate updating the version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,11 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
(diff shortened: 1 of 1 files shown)


djones6:
I believe we need #278 for the tests to pass (consistently) with Swift 5.  I'd like to understand that fix better first though.

djones6:
Rebased branch on top of #278

PR #287, opened 2019-01-17
Use String.split() instead of NSString.components()

ianpartridge:
## Description

Remove the usage of `NSString.components()` and replace it with `String.split()`.

## Motivation and Context
In Swift 5, bridging from `String` to `NSString` involves a conversion from UTF-8 to UTF-16. So we want to avoid `NSString` as much as possible.

## How Has This Been Tested?
Ran Kitura-net test suite.

### DIFF
--- a/Sources/KituraNet/HTTPParser/URLParser.swift
+++ b/Sources/KituraNet/HTTPParser/URLParser.swift
@@ -145,12 +145,11 @@ public class URLParser : CustomStringConvertible {
         }
             
         if let query = query {
-            let pairs = query.components(separatedBy: "&")
+            let pairs = query.split(separator: "&")
             for pair in pairs {
-                    
-                let pairArray = pair.components(separatedBy: "=")
+                let pairArray = pair.split(separator: "=")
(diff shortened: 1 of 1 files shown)


djones6:
Performance evaluation on Ubuntu 16.04 with [a simple query params echo benchmark](https://github.com/IBM-Swift/Kitura-Benchmarks/pull/20):
```
      base_421 |    30848.1 |    31443.5 |    98.2 |        25354 |      4.1 |      4.8 |    201.6 |    10
     split_421 |    31812.4 |    32536.3 |    98.5 |        25715 |      4.0 |      4.8 |    202.6 |    10
       base_50 |    29572.5 |    30033.1 |    98.7 |        26017 |      4.3 |      5.0 |    204.3 |    10
      split_50 |    30675.2 |    31403.9 |    98.3 |        26116 |      4.1 |      4.8 |     50.2 |    10
```
There's a ~3% improvement to the overall performance with this change.  Swift 5 came out behind in this workload however, which implies there may be other NSString usage in the query parameters codepaths to eliminate.

PR #290, opened 2019-01-27
Run Kitura tests in travis CI

pushkarnk:
Run Kitura tests in Travis CI

## Description
After running Kitura-net's test suite, Kitura can be cloned and the Kitura-net dependency updated to the PR branch, enabling the running of Kitura tests.

## Motivation and Context
Many bugs actually escape Kitura-net and Kitura-NIO's tests and are caught in Kitura's tests.

## How Has This Been Tested?
#289 shows a Kitura test failure while Kitura-net tests passed

### DIFF
--- a/.kitura-test.sh
+++ b/.kitura-test.sh
@@ -1,37 +1,19 @@
+set -ex
+
 # Run Kitura-net tests
 swift test
-echo ">> Done running tests. Now preparing to run Kitura tests ..."
 
 # Clone Kitura
-echo ">> cd .. && git clone https://github.com/IBM-Swift/Kitura && cd Kitura"
 cd .. && git clone https://github.com/IBM-Swift/Kitura && cd Kitura
 
 # Build once
-echo ">> swift build"
 swift build
 
 # Edit package Kitura-net to point to the current branch
-echo ">> swift package edit Kitura-net --path ../Kitura-net"
 swift package edit Kitura-net --path ../Kitura-net
-echo ">> swift package edit returned $?."
-
-# If the `swift package edit` command failed, exit with the same failure code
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,11 +14,11 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
-      env: SWIFT_SNAPSHOT=4.0.3
```
djones6: Why are we only applying this to a subset of the builds?  (and how did you choose this subset?)

.kitura-test.sh
```diff
@@ -0,0 +1,37 @@
+# Run Kitura-net tests
+swift test
+echo ">> Done running tests. Now preparing to run Kitura tests ..."
```
djones6: Since this script echoes pretty much every command, could we consider using `set -ex` instead of manually echoing (and manually handling failure)?

The only downside would be that you wouldn't execute the final `cd` if the tests fail, but that probably doesn't matter, since we probably don't care about the linter if the tests fail.

.travis.yml
pushkarnk: Yeah, I just made sure `Linux` and `macOS` are chosen, the selection was random beyond that point. Do you think we should enable it for every build? The only thing I'm worried about in that case is the increased build time.

djones6: Certainly makes sense to include both Mac and Linux.  Having pondered the 'which builds' question, the purpose is primarily to test that we haven't broken the dependent project, and APIs can differ between platforms (but unlikely to differ across Swift versions).  So I'd say we should apply this to the latest release (Swift 5.0 currently) - and maybe that's enough?

PR #292, opened 2019-02-27
Add a query parameters test

pushkarnk:
Add a test to verify query parameters sent to a server
 
## Description
We currently run Kitura tests against every Kitura-net pull request. It makes sense to track bugs that escaped the Kitura-net test suite, but were caught by the Kitura test suite. We must add new tests Kitura-net on every such occurrence. This test was added to Kitura-NIO.

Is it relevant to Kitura-net? Yes. A regressive patch like this one would still go through the Kitura-net tests:

```
--- a/Sources/KituraNet/HTTP/HTTPServerRequest.swift
+++ b/Sources/KituraNet/HTTP/HTTPServerRequest.swift
@@ -152,7 +152,7 @@ public class HTTPServerRequest: ServerRequest {
             Log.error("Host header not received")
         }
         
-        url.append(httpParser?.urlString ?? "")
+        url.append(httpParser?.urlString.removingPercentEncoding ?? "")
```

With the test in this pull request, we'd be able to catch it.

## Motivation and Context
We are incrementally strengthening the Kitura-net/Kitura-NIO test suites every time we find a bug escaping the tests and getting caught in Kitura testing or elsewhere.

### DIFF
--- a/Tests/KituraNetTests/ClientE2ETests.swift
+++ b/Tests/KituraNetTests/ClientE2ETests.swift
@@ -38,7 +38,8 @@ class ClientE2ETests: KituraNetTest {
             ("testPipeliningSpanningPackets", testPipeliningSpanningPackets),
             ("testSimpleHTTPClient", testSimpleHTTPClient),
             ("testUrlURL", testUrlURL),
-            ("testRedirection", testRedirection)
+            ("testRedirection", testRedirection),
+            ("testQueryParameters", testQueryParameters)
         ]
     }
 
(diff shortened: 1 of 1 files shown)


Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -424,6 +425,41 @@ class ClientE2ETests: KituraNetTest {
...
+                }
+                return queryParameters
+            }
+
+            func handle(request: ServerRequest, response: ServerResponse) {
+               do {
+                   let urlComponents = URLComponents(url: request.urlURL, resolvingAgainstBaseURL: false) ?? URLComponents()
+                   let queryParameters = toDictionary(urlComponents.queryItems)
+                   XCTAssertEqual(queryParameters.count, 3, "Expected 3 query parameters, received \(queryParameters.count)")
+                   XCTAssertEqual(queryParameters["key1"], "value1", "Value of key1 should have been value1, received \(queryParameters["key1"]!)")
```
djones6: Rather than the force unwrap, could we have something like:
```swift
received \(queryParameters["key1"] ?? "nil")
```

pushkarnk: Done.

PR #213, opened 2017-07-31
Add support for Unix sockets

adellibovi:
## Description
This PR add a new case for the ClientRequest.Options which is `socketPath(String)`

## Motivation and Context
It resolves issues #212 

## How Has This Been Tested?
NA

## Checklist:
- [x] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)
- [x] If applicable, I have updated the documentation accordingly.
- [x] If applicable, I have added tests to cover my changes.

PR #296, opened 2019-03-25
Add support for Unix sockets

djones6:
## Description

This is a work-in-progress for adding support for Unix domain sockets (client and server side).  This includes a non-breaking version of #213 refactored from @adellibovi's PR, the equivalent server-side changes to allow listening on a Unix socket, and a test.

There is an equivalent PR for Kitura-NIO here: https://github.com/IBM-Swift/Kitura-NIO/pull/187 - naming needs to be reconciled in a few places.

Note: the level of CCurl required to support this is not available on Ubuntu 14.04.  I recently removed 14.04 from the CI for this repo (#295), so the tests will pass, but the Unix socket **client** support will fail (at runtime) on 14.04.  The same limitation may not apply to Kitura-NIO.

## Checklist:

### DIFF
--- a/Sources/KituraNet/ClientRequest.swift
+++ b/Sources/KituraNet/ClientRequest.swift
@@ -148,6 +148,10 @@ public class ClientRequest {
 
     /// Should HTTP/2 protocol be used
     private var useHTTP2 = false
+    
+    /// The Unix socket path used for the request
+    private var socketPath: String? = nil
+
 
     /// Data that represents the "HTTP/2 " header status line prefix
     fileprivate static let Http2StatusLineVersion = "HTTP/2 ".data(using: .utf8)!
@@ -210,6 +214,9 @@ public class ClientRequest {
         /// If present, the client will try to use HTTP/2 protocol for the connection.
         case useHTTP2
         
+        /// If present, the client will connect using the Unix Socket
+        case socketPath(String)
+        
     }
 
     /**
@@ -252,7 +259,7 @@ public class ClientRequest {
         for option in options  {
             switch(option) {
 
-                case .method, .headers, .maxRedirects, .disableSSLVerification, .useHTTP2:
+                case .method, .headers, .maxRedirects, .disableSSLVerification, .useHTTP2, .socketPath:
                     // call set() for Options that do not construct the URL
                     set(option)
                 case .schema(var schema):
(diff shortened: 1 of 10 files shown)


djones6:
@ianpartridge Please could you review? @pushkarnk has an equivalent prototype for Kitura-NIO here: https://github.com/IBM-Swift/Kitura-NIO/pull/187 which seems substantially similar, though there are a couple of naming things to reconcile.

Sources/KituraNet/ClientRequest.swift
```diff
@@ -249,6 +254,10 @@ public class ClientRequest {
         var path = ""
         var port = ""
 
+        if let socketPath = socketPath {
```
ianpartridge: This isn't necessary, we can just `self.socketPath = socketPath`.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -57,7 +57,10 @@ public class HTTPServerRequest: ServerRequest {
     
     /// Client connection socket
     private let socket: Socket
```
ianpartridge: Suggest making this `internal`.

Sources/KituraNet/HTTP/HTTPServerRequest.swift
```diff
@@ -57,7 +57,10 @@ public class HTTPServerRequest: ServerRequest {
...
-    
+
+    /// Socket's signature (required by tests)
+    let socketSignature: Socket.Signature?
```
ianpartridge: Suggest removing this and making `socket` internal instead.

djones6: Whoops - we already have a `public var signature: Socket.Signature?`, so this change can just go

Tests/KituraNetTests/UnixSocketTests.swift
```diff
@@ -0,0 +1,110 @@
...
+    override func tearDown() {
+        doTearDown()
+        // Clean up temporary file
+        removeTemporaryFilePath(socketFilePath)
+    }
+
+    // Generates a unique temporary file path suitable for use as a Unix domain socket.
+    // On Linux, a path is returned within /tmp
+    // On MacOS, a path is returned within /var/folders
+    func uniqueTemporaryFilePath() -> String {
```
ianpartridge: Can we use Foundation on Linux too?

djones6: FileManager.temporaryDirectory wasn't implemented on Linux in Swift 4.0.3

ianpartridge: Ah OK. Separate conversation.. let's leave this as-is for now.

PR #299, opened 2019-05-16
Allow Kitura-Net to only listen on one network address

TaborKelly:
## Description
@billabt was nice enough to add support for this at the BlueSocket layer [here](https://github.com/IBM-Swift/BlueSocket/commit/9d77d6b2126d099f5e56cace9727068f145daaec). However, I really want to use that functionality to fix https://github.com/IBM-Swift/Kitura/issues/1450.

I also have code ready to go for https://github.com/IBM-Swift/Kitura-NIO/pull/205 and Kitura.

## Motivation and Context
This is necessary to close https://github.com/IBM-Swift/Kitura/issues/1450. Should I open another issue for Kitura-Net?

## How Has This Been Tested?
Manual testing by myself. I also modified the test harness to use the new code path.

## Checklist:

### DIFF
--- a/Sources/KituraNet/FastCGI/FastCGIServer.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServer.swift
@@ -56,7 +56,8 @@ public class FastCGIServer: Server {
      */
     public private(set) var port: Int?
 
-    /// Has the same meaning as node in `getaddrinfo()`.
+    /// The address of the network interface to listen on. Defaults to nil, which means this server will listen on all
+    /// interfaces.
     public private(set) var address: String?
 
     /**
@@ -111,7 +112,8 @@ public class FastCGIServer: Server {
      Listens for connections on a socket
      
      - Parameter on: port number for new connections
-     - Parameter address: has the same meaning as node in `getaddrinfo()`
+     - Parameter address: The address of a network interface to listen on, for example "localhost". The default is nil,
+                 which listens for connections on all interfaces.
 
      ### Usage Example: ###
      ````swift
@@ -151,7 +153,8 @@ public class FastCGIServer: Server {
      Static method to create a new `FastCGIServer` and have it listen for conenctions
 
      - Parameter on: port number for accepting new connections
-     - Parameter address: has the same meaning as node in `getaddrinfo()`
+     - Parameter address: The address of a network interface to listen on, for example "localhost". The default is nil,
+                 which listens for connections on all interfaces.
      - Parameter delegate: the delegate handler for FastCGI/HTTP connections
 
      - Returns: a new `FastCGIServer` instance
--- a/Sources/KituraNet/FastCGI/FastCGIServer.swift
+++ b/Sources/KituraNet/FastCGI/FastCGIServer.swift
@@ -56,8 +56,8 @@ public class FastCGIServer: Server {
      */
     public private(set) var port: Int?
 
-    /// Has the same meaning as in `getaddrinfo()`.
-    public private(set) var node: String?
+    /// Has the same meaning as node in `getaddrinfo()`.
+    public private(set) var address: String?
 
     /**
      A server state.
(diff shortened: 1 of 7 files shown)


Sources/KituraNet/FastCGI/FastCGIServer.swift
```diff
@@ -142,6 +147,36 @@ public class FastCGIServer: Server {
         }
     }
 
+    /**
+     Listens for connections on a socket
+     - Parameter on: port number for new connections
+     */
+    @available(*, deprecated, message: "use 'listen(on:node) throws' instead")
+    public func listen(on port: Int) throws {
```
djones6: The addition of the optional function parameter is non-breaking, so I don't think it's necessary to duplicate the functions here.  In the (port:node:delegate:) case below, you'd need to default node to nil.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -151,14 +154,21 @@ public class HTTPServer: Server {
...
-    public func listen(on port: Int) throws {
+    public func listen(on port: Int, node: String?) throws {
         self.port = port
-        try listen(.inet(port))
+        self.node = node
+        try listen(.inet(port, node))
+    }
+
+    @available(*, deprecated, message: "use 'listen(on:node) throws' instead")
+    public func listen(on port: Int) throws {
```
djones6: Same comment as for FastCGIServer - shouldn't need two functions as long as `node` is given a default of nil

Sources/KituraNet/FastCGI/FastCGIServer.swift
TaborKelly: I was trying not to break the public `Server` [protocol](https://github.com/IBM-Swift/Kitura-net/blob/master/Sources/KituraNet/Server/Server.swift) where default initializers are not allowed.

Are you saying that I can have them both (`listen(on)` and `listen(on:node)`) defined in the Server protocol but only `listen(on:node)` with a default initializer in the actual implementation? If so, that would surely be preferable.

TaborKelly:
As for naming, I am happy to change `node` to `address`. I used `node` to match BlueSocket which in turn matches [getaddrinfo](http://man7.org/linux/man-pages/man3/gai_strerror.3.html).

Sources/KituraNet/FastCGI/FastCGIServer.swift
TaborKelly: If I remove `listen(on)` from `HTTPServer` but leave it in the `Server` protocol definition (where I can't have a default initializer) swift isn't happy:
```
$ swift build
/Users/tabor.kelly/work/taborkelly/Kitura-Net/Sources/KituraNet/HTTP/HTTPServer.swift:41:14: error: type 'HTTPServer' does not conform to protocol 'Server'
public class HTTPServer: Server {
             ^
/Users/tabor.kelly/work/taborkelly/Kitura-Net/Sources/KituraNet/Server/Server.swift:45:10: note: protocol requires function 'listen(on:)' with type '(Int) throws -> ()'; do you want to add a stub?
    func listen(on port: Int) throws
```

Please note, this is with the default initializer in HTTPServer.swift:
```
    public func listen(on port: Int, node: String? = nil) throws {
        self.port = port
        self.node = node
        try listen(.inet(port, node))
    }
```

pushkarnk: Hmm. A known work around to this is writing a protocol extension with a method that has the default valued parameter.

```
extension Server {
    func listen(on port: Int, node: String? = nil) {
        listen(on: port, node: node)
    }
}
```

djones6:
@TaborKelly I tried out @pushkarnk's suggestion with the protocol extension here, and it seems to work:
https://github.com/djones6/Kitura-net/commit/46a097c8ed8bca8f6bb63be5d419aba342e6c980

Here's a quick hack up of the corresponding Kitura change to surface the new option:
https://github.com/djones6/Kitura/commit/a5539cf103ea1e3047b23d64b633160b091b4c4f

TaborKelly:
@djones6, @pushkarnk,

Thanks of the help. djones, I used your code almost verbatim. I only fixed up some comments. I think the changed (stripped) whitespace was in your version, and I just copied it. I'm fine with it if you are, but I am also willing to re-add trailing whitespace depending on what you folks want.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -55,12 +55,15 @@ public class HTTPServer: Server {
     /// The TCP port on which this server listens for new connections. If `nil`, this server does not listen on a TCP socket.
     public private(set) var port: Int?
 
+    /// Has the same meaning as node in `getaddrinfo()`.
```
djones6: How about
> The address of the network interface to listen on.  Defaults to `nil`, which means this server will listen on all interfaces.

TaborKelly:
I think that I changed all the documentation comments to match your suggested wording. If we wanted to we could change the word `interfaces` to `addresses` as it would be more correct. Well, at least for Kitura-Net. I suspect the Kitura-NIO is only listening on IPv4 addresses, but I haven't verified it.

PR #302, opened 2019-06-26
tests: use an in-process server in testSimpleHTTPClient

RudraniW:
This fixes a new failure in `ClientE2ETests.testSimpleHTTPClient`.
The `contentType` of the response from "http://www.ibm.com" changed from "text/html" to "text/html; charset=UTF-8". To avoid such failures in the future, an in-process server is used instead of an external server.

### DIFF
--- a/Tests/KituraNetTests/ClientE2ETests.swift
+++ b/Tests/KituraNetTests/ClientE2ETests.swift
@@ -242,8 +242,8 @@ class ClientE2ETests: KituraNetTest {
                 }
             })
             expectation.fulfill()
+        })
     }
-)}
     func testPostRequests() {
         performServerTest(delegate, asyncTasks: { expectation in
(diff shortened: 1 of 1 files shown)


Tests/KituraNetTests/ClientE2ETests.swift
```diff
@@ -214,17 +214,36 @@ class ClientE2ETests: KituraNetTest {
...
+                let contentTypeValue = response?.headers["Content-Type"]
+                XCTAssertNotNil(contentTypeValue, "No ContentType header in response")
+                if let contentType = contentTypeValue {
+                    XCTAssertEqual(contentType, ["text/plain"], "Content-Type header wasn't `text/plain`")
+                }
+            })
+            expectation.fulfill()
     }
-
+)}
```
pushkarnk: Some indentation issues here.

PR #304, opened 2019-07-25
CCurl available only for Linux or macOS

makleso6:
CCurl available only for Linux or macOS. I Use Kitura on iOS platform, so I move CCurl in other scope

## Description

Create flag KITURA_IOS
for Linux or macOS ClientRequest available(no need use CCurl on iOS )
for Linux or macOS HTTP methods available(no need use CCurl on iOS )
## Motivation and Context

Use Kitura WebServer on iOS platform without magic

## How Has This Been Tested?

I think tests no needed

## Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -18,6 +18,7 @@
  **/
 
 import PackageDescription
+import Foundation
 
 var dependencies: [Package.Dependency] = [
     .package(url: "https://github.com/IBM-Swift/LoggerAPI.git", from: "1.7.3"),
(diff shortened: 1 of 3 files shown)


makleso6:
@djones6 I'll try it

makleso6:
@djones6 D  you think that the removal **ClientRequest** in !KITURA_IOS good practice?
Or I need to write ClientRequest iOS class over URLSession.
because tests on iOS does not work :(

djones6:
The Kitura and Kitura-net tests are heavily dependent on `ClientRequest` so yes, if you wanted to be able to run the tests on iOS using Kitura-net then we'd need a replacement.

You could potentially look at using Kitura-NIO instead: Kitura-NIO uses Swift-NIO and doesn't depend on Curl.  To use it, set `KITURA_NIO` when you compile.

PR #307, opened 2019-08-16
[WIP] feat: Ability to limit request size and connection count

djones6:
## Description

This PR implements a `requestSizeLimit` (maximum bytes for a request, _including_ the headers) and `connectionLimit` on total number of concurrent connections.

This is exposed through a new API in Kitura for the user to set these options upon registering a server: https://github.com/IBM-Swift/Kitura/pull/1481

Example usage from Kitura is as follows:
```swift
let connectionPolicy = IncomingSocketOptions(requestSizeLimit: 100, connectionLimit: 3)
Kitura.addHTTPServer(onPort: port, with: router, connectionPolicy: connectionPolicy)
```

If you do not specify either of the parameters, or do not specify a `connectionPolicy` at all, then default values are used.  The defaults are defined in `IncomingSocketOptions` in Kitura-net here: https://github.com/IBM-Swift/Kitura-net/blob/requestLimit/Sources/KituraNet/IncomingSocketOptions.swift
- I didn't think too long about these defaults.  They can be set to whatever we agree is a sensible value, though I'd avoid setting them too low, as we would potentially break existing deployments.

If a client sends a request that exceeds the limit, then they are immediately sent an `HTTP/1.1 413 Request Entity Too Large` and the connection is closed.
If a client attempts to connect, but we are already at the connection limit, they are immediately sent an `HTTP/1.1 503 Service Unavailable` and the connection is closed.

Some issues I'm aware of:
- There are already parameters `keepAlive` and `allowPortReuse` on `Kitura.addHTTPServer` which could be rolled into this new options struct.
- The policing of the read limit is inexact: we read as many bytes as are provided to us in the next read operation, and if that causes us to exceed the read buffer limit, we then fail.
- I didn't reject immediately in the case where the client provided a `Content-Length` (yet). This would save cycles in the case where a client is well-behaved.

## Motivation and Context

See: https://github.com/IBM-Swift/Kitura/issues/1384

## Checklist:


Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
```diff
@@ -236,7 +240,21 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
         }
         
         let bytes = buffer.bytes.assumingMemoryBound(to: Int8.self) + from
+        let readingBody = httpParser.headersComplete
         let (numberParsed, upgrade) = httpParser.execute(bytes, length: length)
+        if (readingBody) {
```
ianpartridge: Don't need parentheses.

ianpartridge: Also, better named `headersAlreadyComplete`? Because the key thing you're finding out is whether the headers were already complete before the parse attempt, yeah?

Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
```diff
@@ -236,7 +240,21 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
...
+            bodyBytesParsed += numberParsed
+        } else if httpParser.headersComplete {
+            // Headers complete while parsing current buffer. Subtract header length so that
+            // we accurately measure the number of bytes of body data we have parsed.
+            bodyBytesParsed += (numberParsed - httpParser.headersLength)
+        }
```
ianpartridge: I would stick an empty

```swift
} else {
  // Do nothing, still parsing headers
}
```

at the end, for completeness.

Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
```diff
@@ -236,7 +240,21 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
...
         
         let bytes = buffer.bytes.assumingMemoryBound(to: Int8.self) + from
+        let readingBody = httpParser.headersComplete
         let (numberParsed, upgrade) = httpParser.execute(bytes, length: length)
+        if (readingBody) {
+            bodyBytesParsed += numberParsed
+        } else if httpParser.headersComplete {
+            // Headers complete while parsing current buffer. Subtract header length so that
+            // we accurately measure the number of bytes of body data we have parsed.
+            bodyBytesParsed += (numberParsed - httpParser.headersLength)
```
ianpartridge: As we are incrementally parsing the request, couldn't we have previously parsed part of the headers, but not all of them?  In that case, don't we need to subtract just the length of the headers we parsed in this attempt, not the length of all the headers? Sorry if I have misunderstood how this works.

Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
ianpartridge: Ideally this would all be a proper state machine via enum cases, but we are where we are.

Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
djones6: Oh good catch, you're totally right - this is incorrect if the headers are parsed via multiple buffers.

Ideally we'd just subtract the amount of the current buffer that was headers, however the HTTP parser doesn't reveal that information, so what we should be doing is accumulating bytes parsed in all cases. In the one-time situation where we completed headers during the current buffer, subtracting the entire header size is then the correct thing to do.

Sources/KituraNet/HTTP/HTTPServer.swift
```diff
@@ -221,6 +228,7 @@ public class HTTPServer: Server {
             }
 
             let socketManager = IncomingSocketManager()
+            socketManager.serverOptions = self.options
```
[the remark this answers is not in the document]
djones6: Good spot, also 'connectionPolicy' was an old naming idea, and the initializers were inconsistent between macOS/Linux. fixed

Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift
```diff
@@ -240,17 +242,25 @@ public class IncomingHTTPSocketProcessor: IncomingSocketProcessor {
...
             bodyBytesParsed += numberParsed
         } else if httpParser.headersComplete {
             // Headers complete while parsing current buffer. Subtract header length so that
-            // we accurately measure the number of bytes of body data we have parsed.
+            // bodyBytesParsed is an accurate measure the number of bytes of body data.
             bodyBytesParsed += (numberParsed - httpParser.headersLength)
+        } else {
+            // Entire buffer represents headers. We will subtract this later once headers
+            // are complete.
+            bodyBytesParsed += numberParsed
```
ianpartridge: I really don't like how we have something called `bodyBytesParsed` which includes the number of header bytes until we've finished parsing headers... Can we split this up into `requestBytesParsed` and a separate `bodyBytesParsed`?

djones6: I agree, however, this is private to the class and the code comments do adequately cover the situations under which this makes sense.  In the interests of getting this feature implemented I'd vote we leave as-is.

ianpartridge: OK.

PR #321, opened 2021-04-08
Now using modern unsafe syntax; Add BufferListTests from Kitura-NIO

dannys42:
- BufferList: update fill() to make better use of bufferPointer
- ClientRequest: update use of unsafeBufferPointer
- FastCGIRecordParser: update with new unsafe syntax
- HTTPServerResponse: update to use new address syntax
- HTTPParser: update to use new unsafe syntax
- IncomingSocketProcessor: conform to AnyObject rather than class
- ServerDelegate: conform to AnyObject rather than class
- ServerRequest: conform to AnyObject rather than class
- ServerResponse: conform to AnyObject rather than class
- BufferListTests: Add tests from Kitura-NIO
- FastCGIRequestTests: Update copyUInt16IntoBuffer() and copyUInt32IntoBuffer() for clarity rather than performance

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,6 @@ matrix:
       dist: xenial
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=swift:4.0.3 SWIFT_SNAPSHOT=4.0.3
-    - os: linux
-      dist: xenial
-      sudo: required
-      services: docker
       env: DOCKER_IMAGE=swift:4.1.3 SWIFT_SNAPSHOT=4.1.3
     - os: linux
       dist: xenial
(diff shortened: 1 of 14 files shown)


dannys42:
@mbarnach Any thoughts on this PR?  Some of it is a copy from the recent Kitura-NIO changes.  But I think a lot of it is just updating how we manage endian with more modern Swift syntax.  I hope I got everything right ;-)   But I think macOS and Linux are both little-endian.  So if we wanted to be sure, someone would probably have to check this on an ARM or something.  hmm... I wonder if we can add that to our travis builds?

mbarnach (approved):
👍

PR #322, opened 2021-07-16
Removed Swift 4.x from CI; updated IBM-Swift org to Kitura; update jazzy docs

dannys42:


dannys42:
@mbarnach Have any thoughts on why this is failing?  Maybe something changed in Swift 5.4?

mbarnach:
🤔 That's weird? I've you tried to re-run this build only? It seems random, but the test is not super clear neither (13 or 16 bytes?).
I don't think it is Swift 5.4 related.

